PR #5, opened 2019-09-01
available in other repositories

sadayuki-matsuno:
Changed to be available in other repositories.

Since rev-parse returns an error when there is no commit, it is changed to rev-list.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -66,8 +66,18 @@ func clone(url, repoPath string) error {
 func pull(repoPath string) ([]string, error) {
 	commandArgs := generateGitArgs(repoPath)
 
+	remoteCmd := []string{"remote", "get-url", "--push", "origin"}
+	output, err := utils.Exec("git", append(commandArgs, remoteCmd...))
+	if err != nil {
+		return nil, xerrors.Errorf("error in git rev-list: %w", err)
+	}
+	remoteURL := strings.TrimSpace(output)
+	if remoteURL != repoPath {
+		return nil, xerrors.Errorf("remote url is %s", remoteURL)
+	}
+
 	revParseCmd := []string{"rev-list", "-n", "1", "--all"}
-	output, err := utils.Exec("git", append(commandArgs, revParseCmd...))
+	output, err = utils.Exec("git", append(commandArgs, revParseCmd...))
 	if err != nil {
 		return nil, xerrors.Errorf("error in git rev-list: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -41,9 +43,17 @@ func run() error {
 	flag.Parse()
 	now := time.Now().UTC()
 
+	repoOwner := os.Getenv("VULNLIST_REPOSITORY_OWNER")
```
knqyf263: It seems that LookupEnv is better. In addition, this process should be functionalized like utils.LookupEnv(key, defaultValue string).

git/git.go
```diff
@@ -66,12 +66,16 @@ func clone(url, repoPath string) error {
...
-	revParseCmd := []string{"rev-parse", "HEAD"}
+	revParseCmd := []string{"rev-list", "-n", "1", "--all"}
 	output, err := utils.Exec("git", append(commandArgs, revParseCmd...))
 	if err != nil {
-		return nil, xerrors.Errorf("error in git rev-parse: %w", err)
+		return nil, xerrors.Errorf("error in git rev-list: %w", err)
 	}
 	commitHash := strings.TrimSpace(output)
+	if len(commitHash) == 0 {
+		log.Println("no commit yet")
```
knqyf263: Use logger package.

main.go
sadayuki-matsuno: fixed

main.go
```diff
@@ -41,10 +43,19 @@ func run() error {
...
+	if val, ok := os.LookupEnv("VULNLIST_REPOSITORY_NAME"); ok {
+		repoName = val
+	}
+
 	// Embed GitHub token to URL
 	githubToken := os.Getenv("GITHUB_TOKEN")
-	url := fmt.Sprintf(repoURL, githubToken)
+	url := fmt.Sprintf(repoURL, githubToken, repoOwner, repoName)
 
+	log.Printf("target repository is %s/%s", repoOwner, repoName)
```
knqyf263:
```suggestion
	log.Printf("target repository is github.com/%s/%s\n", repoOwner, repoName)
```

main.go
knqyf263: Would you create new function such as `utils.LookupEnv(key, defaultValue string)`?

knqyf263: func LookupEnv(key, defaultValue string){
    if val, ok := os.LookupEnv(key); ok {
		return val
    }
    return defaultValue
}

PR #9, opened 2019-10-02
Support Amazon Linux AMI Security Advisory

knqyf263:


amazon/amazon_test.go
```diff
@@ -0,0 +1,135 @@
...
+	testCases := []struct {
+		name          string
+		xmlResponse   string
+		gzipFileName  string
+		expectedError error
+	}{
+		{
+			name:          "happy path",
+			xmlResponse:   goodRepoMDXMLResponse,
+			gzipFileName:  "test-data/updateinfo.xml.gz.valid",
```
knqyf263:
```suggestion
			gzipFileName:  "testdata/updateinfo.xml.gz.valid",
```
I think `testdata` is popular.
https://github.com/golang/go/tree/master/src/cmd/go/testdata

knqyf263: In addition, haven't you pushed `testdata` yet?

simar7: Yeah sorry I forgot to push that, but I believe you've already added this so I'll resolve.

simar7:
Should we merge this in @knqyf263 ?

amazon/amazon_test.go
```diff
@@ -0,0 +1,115 @@
+package amazon_test
```
simar7: Could we keep the test files in the same package? It helps with showing test coverage within a package and also allows for testing private functions that otherwise won't be able to get tested if it was in a different package.

knqyf263: First of all, this is a well-known technique. Many Go standard packages use this technique.
https://github.com/golang/sys/blob/master/unix/mmap_unix_test.go

I think this has some advantages.
https://segment.com/blog/5-advanced-testing-techniques-in-go/

The biggest advantage is we have a chance to think if an interface is appropriate and easy-to-use.

simar7: I think I'm willing to try this new approach after reading the article! Thanks for sharing!

knqyf263: My pleasure 👍

knqyf263:
@simar7 Thanks!

PR #12, opened 2019-10-06
Introduce GitHub Actions

knqyf263:


.github/workflows/go.yml
```diff
@@ -0,0 +1,29 @@
...
+      run: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s $version
+      env:
+        version: v1.19.1
+
+    - name: Lint
+      run: ./bin/golangci-lint run
+          
+    - name: Test
+      run: |
+        go test ./...
```
simar7: Might want to add the `-v` flag for more verbose output.

simar7:
Other than the nit, it lgtm!

.github/workflows/go.yml
knqyf263: Thanks. I added.

simar7:
Hey @knqyf263 this includes a revert commit: https://github.com/aquasecurity/vuln-list-update/pull/12/commits/71637ba93e2e04434526d0675dbbe2b5ba416ff7 which was needed since the underlying mock library uses locks which in turn would result in us changing all our method receivers in various places. 


A simple example of getting in trouble with passing locks by value would be the following:
```
package main
import "sync"
type T struct {
    lock sync.Mutex
}
func (t *T) Lock() {
    t.lock.Lock()
}
func (t T) Unlock() {
   t.lock.Unlock()
}
func main() {
    t := T{lock: sync.Mutex{}}
    t.Lock()
    t.Unlock()
    t.Lock()
}
```

This program will deadlock.

I realize maybe we need a separate PR to handle that but without the revert commit the GitHub actions will still be red.

knqyf263:
Hi @simar7, I think it's easy to fix it. We should make the receiver a pointer instead of a value. Now, all tests have passed.

knqyf263:
@simar7 Thanks a lot!

PR #13, opened 2019-10-07
refactor(alpine): remove unused functions and add tests

knqyf263:
Alpine transferred their ticket tracker from Redmine to GitLab. So the Redmine specific code should be removed.
https://gitlab.alpinelinux.org/alpine/aports


simar7:
Sorry but I missed a commit in this PR: https://github.com/aquasecurity/vuln-list-update/pull/13/commits/3e42145b06db57657d6eb8256beb82beb001867b I have made some comments in it. I will try to make another PR with the fixes.

alpine/export_test.go
```diff
@@ -0,0 +1,9 @@
+package alpine
```
simar7: Having global state is an anti-pattern, I'm going to cut a new PR out to take care of this.

main.go
```diff
@@ -42,7 +42,8 @@ func main() {
 func run() error {
 	flag.Parse()
 	now := time.Now().UTC()
-	gc := git.Config{}
+	gc := &git.Config{}
```
simar7: Why do we need this to pass this by reference?

alpine/alpine_test.go
```diff
@@ -1,43 +1,42 @@
-package alpine
+package alpine_test
```
simar7: I like the idea of keeping the tests in the same package as the package itself. One advantage I see with it is not having an extra import for a package and being able to test un-exported stuff inside.

alpine/alpine_test.go
```diff
@@ -1,43 +1,42 @@
...
+	mock.Mock
 }
 
 func (mgc MockGitConfig) CloneOrPull(a string, b string) (map[string]struct{}, error) {
-	if mgc.cloneorpull != nil {
-		return mgc.cloneorpull(a, b)
-	}
-	return map[string]struct{}{}, nil
+	args := mgc.Called(a, b)
+	return args.Get(0).(map[string]struct{}), args.Error(1)
```
simar7: I would propose against this change. The author advises that this is prone to panic (rightfully so) here: https://godoc.org/github.com/stretchr/testify/mock and having extra checks to see if the object is nil or not seems bit of an overkill.

alpine/alpine_test.go
knqyf263: At first, this is a well-known technique. Many Go standard packages use this technique.
https://github.com/golang/sys/blob/master/unix/mmap_unix_test.go

I think this has some advantages.
https://segment.com/blog/5-advanced-testing-techniques-in-go/

The biggest advantage is we have a chance to think if an interface is appropriate and easy-to-use.

main.go
knqyf263: I think either is fine, but it is better to reduce the use of memory.

alpine/export_test.go
knqyf263: No. In the case of testing, `export_test.go` is a well-known technique.
https://github.com/golang/go/blob/master/src/bufio/export_test.go

You can see a lot of examples here.
https://github.com/golang/go/search?q=filename%3Aexport_test.go&unscoped_q=filename%3Aexport_test.go

*_test.go can not be referred from standard files. In short, these exported functions and variables are available only in testing.

Issue #10, opened 2019-10-03
Rejected vulnerabilities


knqyf263:
`vuln-list-update` adds and updates vulnerabilities. However, in rare cases, vulnerabilities may be removed like a rejected one.
e.g. https://security-tracker.debian.org/tracker/CVE-2019-9619

This should be removed from vuln-list.
https://github.com/aquasecurity/vuln-list/blob/7644605af8bf37b38b437efd6377d1002df606f7/oval/debian/buster/2019/CVE-2019-9619.json

PR #15, opened 2019-10-15
fix(debian): remove rejected vulnerabilities

knqyf263:
Close https://github.com/aquasecurity/vuln-list-update/issues/10

### DIFF
--- a/debian/debian_test.go
+++ b/debian/debian_test.go
@@ -81,7 +81,7 @@ func TestClient_Update(t *testing.T) {
 				_ = os.Mkdir(d, 0777)
 				for _, fileName := range tc.existedFiles {
 					err = ioutil.WriteFile(filepath.Join(d, fileName), []byte("test"), 0666)
-					assert.Nil(t, err)
+					assert.Nil(t, err, "failed to write the file")
 				}
 			}
 
@@ -98,7 +98,6 @@ func TestClient_Update(t *testing.T) {
 			if tc.expectedError == "" {
 				assert.Nil(t, err)
 			} else {
-				fmt.Println(err)
 				assert.Contains(t, err.Error(), tc.expectedError)
 				return
 			}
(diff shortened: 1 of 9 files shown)


debian/debian_test.go
```diff
@@ -95,11 +95,11 @@ func TestClient_Update(t *testing.T) {
...
-			if tc.expectedError == "" {
-				assert.Nil(t, err)
-			} else {
-				assert.Contains(t, err.Error(), tc.expectedError)
-				return
+			switch {
+			case tc.expectedError != "":
+				assert.Contains(t, err.Error(), tc.expectedError, tc.name)
+			default:
+				assert.NoError(t, err, tc.name)
```
knqyf263: Why did you replace `if` with `switch`?

debian/debian_test.go
```diff
@@ -72,7 +72,7 @@ func TestClient_Update(t *testing.T) {
 
 			fmt.Println(path.Join(testServer.URL, tc.path))
 			dir, err := ioutil.TempDir("", "debian")
-			assert.Nil(t, err, "failed to create temp dir")
+			assert.NoError(t, err, "failed to create temp dir")
```
knqyf263: Thanks!

debian/debian_test.go
simar7: Just personal preference. Looks a bit cleaner IMO.

knqyf263: I see.

debian/debian_test.go
```diff
@@ -0,0 +1,133 @@
...
+			err = client.Update()
+			switch {
+			case tc.expectedError != "":
+				assert.Contains(t, err.Error(), tc.expectedError, tc.name)
+			default:
+				assert.NoError(t, err, tc.name)
+			}
+
+			// Compare got and golden
+			err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
simar7: I understand what this check is doing but IMO this feels like a need to have the utils package expose an interface that others can implement (especially in a test). At the moment, this would be an integration test that touches actual files on disk rather than a unit test (self contained). I added a TODO for us to revisit this in the future.

knqyf263: I understand what you mean well, but I feel that the abstraction of file I/O is a little too much. It makes the code complex. 
As you said, this is not strictly a unit test. We should discuss it later.

PR #19, opened 2019-11-03
feat(rhel_oval): support RHEL OVAL

knqyf263:
Recently, Red Hat Security Data API changed a behavior. They don't return epoch of a patched version. 
https://access.redhat.com/documentation/en-us/red_hat_security_data_api/1.0/html/red_hat_security_data_api/cve

We have to use OVAL (or CVRF) for patched vulnerabilities now.
https://www.redhat.com/security/data/oval/v2/


oval/redhat/redhat_test.go
```diff
@@ -0,0 +1,146 @@
+package redhat
```
simar7: Is this supposed to be a file with integration tests? If so we should start the habit of using build tags `// +build integration` to avoid long running tests.

oval/redhat/redhat_test.go
```diff
@@ -0,0 +1,146 @@
...
+				expected, err := ioutil.ReadFile(goldenPath)
+				assert.NoError(t, err, tc.name)
+
+				assert.Equal(t, expected, actual, tc.name)
+
+				return nil
+			})
+			assert.Equal(t, len(tc.goldenFiles), fileCount, tc.name)
+			assert.NoError(t, err, tc.name)
+		})
```
simar7: Can we factor this out to a separate integration test file? Keeping a unit test (the first half of this test) and the integration test together is confusing to read and hard to maintain.

oval/redhat/redhat.go
```diff
@@ -0,0 +1,110 @@
...
+	bar.Finish()
+	return nil
+}
+
+func (c Config) saveRHSAPerYear(dirName string, rhsaID string, data interface{}) error {
+	// e.g. RHSA-2018:0094
+	s := strings.Split(rhsaID, ":")
+	if len(s) != 2 {
+		log.Printf("invalid RHSA-ID format: %s\n", rhsaID)
+		return nil
```
simar7: Why do we return `nil` back if the format was invalid?

oval/redhat/redhat.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	s = strings.Split(s[0], "-")
+	if len(s) != 2 {
+		log.Printf("invalid RHSA-ID format: %s\n", rhsaID)
+		return nil
```
simar7: Why do we return `nil` back if the format was invalid?

oval/redhat/redhat.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		bar.Increment()
+	}
+	bar.Finish()
+	return nil
+}
+
+func (c Config) saveRHSAPerYear(dirName string, rhsaID string, data interface{}) error {
+	// e.g. RHSA-2018:0094
+	s := strings.Split(rhsaID, ":")
```
simar7: This might be better suited for a regex if the format remains the same. Doing a `split` over delimiters is fragile. This should do the trick: `(RHSA)\-\d{4}\:\d{4}`

oval/redhat/redhat.go
simar7: I've added a failing test to showcase the behaviour seen: https://github.com/aquasecurity/vuln-list-update/pull/19/commits/39257bba851c947404937720b046128ba37bfed6

oval/redhat/types.go
```diff
@@ -0,0 +1,108 @@
+package redhat
```
simar7: I'm curious is this a generated file? If so, how was it generated?

oval/redhat/types.go
```diff
@@ -0,0 +1,108 @@
...
+	AffectedCPEList []string   `xml:"affected_cpe_list>cpe"`
+	Issued          struct {
+		Date string `xml:"date,attr"`
+	} `xml:"issued"`
+	Updated struct {
+		Date string `xml:"date,attr"`
+	} `xml:"updated"`
+}
+
+// Cve : >definitions>definition>metadata>advisory>cve
```
simar7: Also curious, what do these comments signify?

oval/redhat/redhat_test.go
knqyf263: This file is for unit testing. (though it uses httptest...). At least, this test finishes in 0.1 sec.

```
ok      github.com/aquasecurity/vuln-list-update/oval/redhat    0.039s
```

knqyf263: I think we don't need a tag yet.

oval/redhat/redhat.go
knqyf263: `afero` is used for FileSystem mocking. If you use NewMemMapFs, files are saved on a memory, not on a local file system. So, this is unit testing.
https://github.com/spf13/afero#step-2-declare-a-backend

oval/redhat/redhat_test.go
knqyf263: afero is used for FileSystem mocking. If you use NewMemMapFs, files are saved on a memory, not on a local file system. So, this is unit testing.
https://github.com/spf13/afero#step-2-declare-a-backend

oval/redhat/redhat.go
knqyf263: Red Hat data is sometimes broken. In that case, if we return err, the entire update will be stopped. For example, the Red Hat xml has 1,000 vulnerabilities. Even if 1 vulnerability is broken, 999 vulnerabilities should be saved. If we return err, the update of 1,000 vulnerabilities will fail.

oval/redhat/redhat.go
knqyf263: Ditto

oval/redhat/redhat.go
knqyf263: A regexp of Go is not fast, so I always avoid using it except for complicated patterns. Also, a regexp is often hard to understand. Therefore I used `split`, but I don't care about it. If you think we should use a regexp, I can replace this one with a regexp.

oval/redhat/types.go
knqyf263: No, it's manual. I tried some tools, but they don't work as I expected.
https://www.onlinetool.io/xmltogo/

oval/redhat/types.go
knqyf263: This just shows the structure of XML. OVAL XML has a deep nest and confuses me. 
definitions
  -> definition
        -> metadata...

oval/redhat/redhat_test.go
```diff
@@ -142,5 +143,44 @@ func TestConfig_Update(t *testing.T) {
 			assert.NoError(t, err, tc.name)
 		})
 	}
+}
+
+func TestConfig_saveRHSAPerYear(t *testing.T) {
```
knqyf263: I've already written tests for `Update`. At the same time, we can test the behavior of `saveRHSAPerYear` through `Update`. These test cases are already tested in `TestConfig_Update`. This method is private. Basically, I think we don't have to test private methods. It is sometimes difficult to test a private method via a public method. Or, we sometimes want to test many cases. In that case, I write unit tests even for a private method. This is a little famous religious war...

simar7: My 2¢: I fully agree with testing exported methods first and I also believe in testing unexported methods through the means of calling exported methods that in turn call the former. 

But if calling an exported method will never invoke the unexported method in question under a test environment, then the unexported method should be removed. It is essentially dead code at that point that will never see the light of the day in real usage. Of course this is all based on the premise that we're writing tests as close as possible to the real environment. 

By definition, a unit of testing in a unit test is a method. If we're not testing each method, exported or not, we're not testing each unit independently.

oval/redhat/redhat.go
simar7: Fair point. Let's keep it this way then.

oval/redhat/redhat.go
simar7: Hmm I see. I think in that case we should at least distinguish between a good return `return nil` vs a bad return `return something else` in order clarify to the caller what happened. As it stands there's no way to test if the call to `saveRHSAPerYear` was a success or a failure because return value is the same for both.

simar7: I've added a commit to help that: https://github.com/aquasecurity/vuln-list-update/pull/19/commits/b8e8adf4966f1c05a2f839f9fefc88f3c81b1b62

oval/redhat/redhat_test.go
knqyf263: I completely agree with your idea that uncalled functions should be removed.

>If we're not testing each method, exported or not, we're not testing each unit independently.
So, does it mean that we should also test private methods based on unit test definitions?

A unit test is useful for refactoring. If there is a unit test, we can change the internal code without changing the behavior from the outside. But if we write a test for unexported functions, it sometimes prevents refactoring. This is because that test is for implementation, not for behavior. This is not an absolute rule. We should write a test for an unexported function sometimes.

simar7: Agreed! On that point, if something's hard to unit test it should be refactored out. In this case it seems the without actually refactoring the unexported method, it will be hard to test through the exported one.

knqyf263: Hmm... I didn't feel hard to test in this case because I just divided one function into the exported one and the unexported one. Which points do you feel hard? How should we improve?

simar7: Without having a test for `saveRHSAPerYear` this codepath is untested: https://github.com/aquasecurity/vuln-list-update/pull/19/files#diff-ceaa4154ecb6fbceb0b0479e253bbdb6R99-R102

knqyf263:
@simar7 thanks!

oval/redhat/redhat_test.go
knqyf263: I understand. I've already passed the XML including an invalid RHSA-ID in TestConfig_Update. So this codepath (https://github.com/aquasecurity/vuln-list-update/commit/b8e8adf4966f1c05a2f839f9fefc88f3c81b1b62#diff-ceaa4154ecb6fbceb0b0479e253bbdb6R103-R106) is already tested, but that codepath was not tested. Although I knew it, I thought it was too much. In this case, as you said, it might have been better to test it. Thanks.

PR #24, opened 2019-12-07
Add photon advisory

masahiro331:
- [x] Add photon advisory support 
- [x] Add tests


main.go
```diff
@@ -139,6 +140,12 @@ func run() error {
 			return xerrors.Errorf("error in SUSE CVRF update: %w", err)
 		}
 		commitMsg = "SUSE CVRF"
+	case "photon":
+		op := photon.NewConfig()
```
knqyf263: op?

photon/photon.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (c Config) Update() error {
+	log.Printf("Fetching Photon")
+
+	versions, err := c.getPhotonVersion()
+	if err != nil {
+		return xerrors.Errorf("failed to fetch Photon versions: %w", err)
+	}
+	for _, version := range versions {
```
knqyf263: recommend: Add a log message like log.Printf("Fetching Photon %s", version)

photon/photon.go
```diff
@@ -0,0 +1,105 @@
...
+		bar.Finish()
+	}
+
+	return nil
+}
+
+func (c Config) saveCVEPerYear(dirName string, cveID string, data interface{}) error {
+	s := strings.Split(cveID, "-")
+	if len(s) != 3 {
+		return xerrors.Errorf("invalid CVE-ID format: %s\n", cveID)
```
knqyf263:
```suggestion
		return xerrors.Errorf("invalid CVE-ID format: %s", cveID)
```

photon/photon.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	yearDir := filepath.Join(c.VulnListDir, dirName, s[1])
+	if err := c.AppFs.MkdirAll(yearDir, os.ModePerm); err != nil {
+		return err
+	}
+
+	filePath := filepath.Join(yearDir, fmt.Sprintf("%s.json", cveID))
```
knqyf263: critical: This overwrites advisories with the same CVE-ID. For example, CVE-2019-9857 has 4 packages such as `linux`, `linux-aws` and `linux-secure`. But only `linux-secure` is saved for now because other 3 packages are overwritten.

photon/photon.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (c Config) saveCVEPerYear(dirName string, cveID string, data interface{}) error {
+	s := strings.Split(cveID, "-")
+	if len(s) != 3 {
+		return xerrors.Errorf("invalid CVE-ID format: %s\n", cveID)
+	}
+
+	yearDir := filepath.Join(c.VulnListDir, dirName, s[1])
+	if err := c.AppFs.MkdirAll(yearDir, os.ModePerm); err != nil {
+		return err
```
knqyf263: wrap

PR #26, opened 2020-01-09
Use shallow clone

Vesnica:
Use shallow clone can reduce the download size by 80% when getting vuln-list repo


knqyf263:
Thanks.

PR #28, opened 2020-01-28
fix(amazon): handle removed ALAS

knqyf263:
Amazon added ALAS-2020-1383, but it was removed now.
https://alas.aws.amazon.com/alas2.html

This sometimes seems to happen. To handle it, we should remove the directory once and add all files to the directory again.

### DIFF
--- a/amazon/amazon.go
+++ b/amazon/amazon.go
@@ -104,6 +104,14 @@ func (ac Config) Update() error {
 }
 
 func (ac Config) update(version, url string) error {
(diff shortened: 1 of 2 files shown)


simar7:
Curious if this is possible with other distros as well @knqyf263?

knqyf263:
@simar7 Yes, I actually fixed the same issue in Debian. We should check other distributions after this PR.
https://github.com/aquasecurity/vuln-list-update/pull/15

PR #29, opened 2020-02-01
Add GitHub Security Advisory Database 

masahiro331:
This Pull Request adds vulnerability information of github security advisory database to vuln-list.

The following Ecosystems are supported.
ref: https://developer.github.com/v4/enum/securityadvisoryecosystem/
- Composer
- Maven
- Npm
- Nuget
- Pip
- Rubygems

The data schema of trivy-db is undecided now.   
If you want to try of GraphQL API can be executed on this site.

https://developer.github.com/v4/explorer/

### Issues: 
- GitHub Security Advisory sometimes does not have CVE-ID
Therefore, it is stored using GHSA-ID.


ghsa/ghsa.go
```diff
@@ -0,0 +1,178 @@
...
+			break
+		}
+
+		variables["cursor"] = githubql.NewString(getVulnerabilitiesQuery.PageInfo.EndCursor)
+	}
+
+	return ghsas, nil
+}
+
+func (c Config) saveGSA(dirName string, ghsaID, data interface{}) error {
```
simar7: Can we make `ghsaID` and `data` as their respective data types rather than an `interface` type?

ghsa/ghsa.go
```diff
@@ -0,0 +1,178 @@
...
+		"total":     graphql.Int(100),
+		"cursor":    (*githubql.String)(nil),
+	}
+	for {
+		var err error
+		for i := 0; i <= c.Retry; i++ {
+			if i > 0 {
+				wait := math.Pow(float64(i), 2) + float64(utils.RandInt()%10)
+				log.Printf("retry after %f seconds\n", wait)
+				time.Sleep(time.Duration(time.Duration(wait) * time.Second))
```
simar7: This logic is currently untested, we could add a mock that fails on first try and passes subsequently.

ghsa/ghsa.go
```diff
@@ -0,0 +1,178 @@
...
+	bar.Finish()
+	return nil
+}
+
+func (c Config) FetchGithubSecurityAdvisories(ecosystem SecurityAdvisoryEcosystem) ([]GithubSecurityAdvisory, error) {
+	var getVulnerabilitiesQuery GetVulnerabilitiesQuery
+	var ghsas []GithubSecurityAdvisory
+	variables := map[string]interface{}{
+		"ecosystem": ecosystem,
+		"total":     graphql.Int(100),
```
simar7: Why is the count of `100` important? Can we increase that?

main.go
```diff
@@ -146,6 +150,17 @@ func run() error {
 			return xerrors.Errorf("error in Photon update: %w", err)
 		}
 		commitMsg = "Photon Security Advisories"
+	case "ghsa":
+		src := oauth2.StaticTokenSource(
+			&oauth2.Token{AccessToken: githubToken},
```
simar7: I'm assuming this GitHub token will actually require the same level of permissions that a token to access the GraphQL api would?

ghsa/ghsa.go
masahiro331: Sorry miss.

ghsa/ghsa.go
masahiro331: The response size of security vulnerability is up to 100.

define const MaxResponseSize 100.

main.go
masahiro331: graphql api does not require any scope...
Should I use another access token?

masahiro331:
@simar7 
Thanks for the fix.
Consultation.
I think Ecosystem is better than Ecosystems for Config members.

For example,
FetchGithubSecurityAdvisories have  ecosystems of Config, but is designed to recieve the ecosystem as an argument.

PR #32, opened 2020-04-05
fix(git): fetch all branches

knqyf263:
In the repository which is cloned with `--depth 1` option, `git branch --remote` doesn't work. Before that, we need the following commands to fetch remote branches.

```
$ git config "remote.origin.fetch", "+refs/heads/*:refs/remotes/origin/*
$ git fetch --all
```

This bug was introduced by https://github.com/aquasecurity/vuln-list-update/pull/26.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -46,6 +46,9 @@ func (gc Config) CloneOrPull(url, repoPath string) (map[string]struct{}, error) 
 			return nil, err
 		}
 
+		if err := fetchAll(repoPath); err != nil {
+			return nil, err
+		}
 		err = filepath.Walk(repoPath, func(path string, info os.FileInfo, err error) error {
 			if info.IsDir() {
 				return nil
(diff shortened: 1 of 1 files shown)


git/git.go
```diff
@@ -115,6 +118,20 @@ func pull(url, repoPath string) ([]string, error) {
 	return updatedFiles, nil
 }
 
+func fetchAll(repoPath string) error {
+	commandArgs := generateGitArgs(repoPath)
+	configCmd := []string{"config", "remote.origin.fetch", "+refs/heads/*:refs/remotes/origin/*"}
+	if _, err := utils.Exec("git", append(commandArgs, configCmd...)); err != nil {
+		return xerrors.Errorf("error in git config: %w", err)
```
danielpacak, line 125: Do we still need xerrors in Go 1.13?
```suggestion
		return fmt.Errorf("error in git config: %w", err)
```

knqyf263: `errors` supported `Wrap`, but unfortunately it doesn't support stack trace yet.
https://play.golang.org/p/xwGyjc6FyRL

This stack trace is useful for debugging, especially when someone raises an issue with the error message. So, I prefer using `xerrors` yet.

danielpacak: Ah I see. Yup it makes sens.

Issue #36, opened 2020-07-13
Migrate to GitHub Actions


knqyf263:
Now, vuln-list-update uses TravisCI. We'll migrate CI to GitHub Actions.
https://github.com/aquasecurity/vuln-list-update/blob/master/.travis.yml

knqyf263:
@rahul2393

rahul2393:
@knqyf263 Manual workflow working and pushing to repo [here](https://github.com/rahul2393/vuln-list)
https://github.com/rahul2393/vuln-list-update/runs/865003803?check_suite_focus=true

PR #37, opened 2020-07-13
Removed travis file and added github workflow

rahul2393:
[Issue](https://github.com/aquasecurity/vuln-list-update/issues/36)
Add secret OWNER for vuln-list repo owner, and ACCESS_TOKEN with rights to push to repo in 
https://github.com/aquasecurity/vuln-list-update/settings/secrets


.github/workflows/update.yml
```diff
@@ -0,0 +1,152 @@
...
+      uses: actions/checkout@v1
+
+    - name: Setup github user email and name
+      run: |
+        git config --global user.email "action@github.com"
+        git config --global user.name "GitHub Action"
+        
+    - name: Run NVD
+      env:
+        VULNLIST_REPOSITORY_OWNER: ${{ secrets.Owner }}
```
knqyf263, line 24: If we don't specify it, `aquasecurity` will be used. Do we need this one?
https://github.com/aquasecurity/vuln-list-update/blob/11f2a4cbee7fe480dbfe65972784d61ac98aad62/main.go#L56

.github/workflows/update.yml
```diff
@@ -0,0 +1,152 @@
...
+        GITHUB_TOKEN: ${{ secrets.ACCESS_TOKEN }}
```
knqyf263, line 25: In addition, can we use env in job? I'd like to avoid writing `env` for all steps.
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idenv

.github/workflows/update.yml
```diff
@@ -0,0 +1,152 @@
...
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
+
+    - name: Setup github user email and name
+      run: |
+        git config --global user.email "action@github.com"
+        git config --global user.name "GitHub Action"
+        
+    - name: Run NVD
```
knqyf263, line 22: nit: other steps don't have `Run`.

```suggestion
    - name: NVD
```

.github/workflows/update.yml
```diff
@@ -0,0 +1,152 @@
...
+
+jobs:
+  update:
+    name: Update repo vuln-list
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
```
knqyf263, line 12: Let's bump up Go to 1.14.
```suggestion
    - name: Set up Go 1.14
      uses: actions/setup-go@v1
      with:
        go-version: 1.14
```

.github/workflows/update.yml
rahul2393: No not needed, we can remove it, needed in my case to push to fork repo

.github/workflows/update.yml
```diff
@@ -0,0 +1,89 @@
...
+      VULNLIST_REPOSITORY_OWNER: $(echo "$GITHUB_REPOSITORY" | awk -F / '{print $1}')
+      GITHUB_TOKEN: ${{ secrets.ACCESS_TOKEN }}
+    steps:
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.14
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
```
knqyf263, line 18:
```suggestion
      uses: actions/checkout@v2
```

.github/workflows/update.yml
```diff
@@ -0,0 +1,89 @@
...
+
+    - name: Setup github user email and name
+      run: |
+        git config --global user.email "action@github.com"
+        git config --global user.name "GitHub Action"
+
+    - name: NVD
+      run: go run main.go -target nvd
```
knqyf263, line 26: These steps are stateful, so we might be able to compile the binary and use it in later steps.

```suggestion
    - name: Compible vuln-list-update
      run: go install

   - name: NVD
      run: vuln-list-update -target nvd
```

or

```suggestion
    - name: Compible vuln-list-update
      run: go build -o vuln-list-update

   - name: NVD
      run: ./vuln-list-update -target nvd
```

PR #38, opened 2020-07-13
Update to use github cron workflow

rahul2393:


knqyf263:
Related issue: https://github.com/aquasecurity/vuln-list-update/pull/37

PR #40, opened 2020-07-29
cwe: Add initial logic to download and save CWE files

simar7:
This PR adds the ability to save CWE information as part of the vuln-list project adding another layer of information about each vulnerability.

It will save 2 files: `cwe.xml` and `cwe.json`. The idea is to use `cwe.json` as a data document where the `ID` will be the index key to gather information for each CWE.

Signed-off-by: Simarpreet Singh <simar@linux.com>


simar7:
Maybe we don't need the `cwe.xml` after all since we will have `cwe.json` to work with. What do you think @knqyf263?

knqyf263:
I also don't think `cwe.xml` is necessary in `vuln-list`. Other sources don't store the original data.
Also, what if we store CWE-IDs separately such as CWE-79.json? It enables us to track the data easily. If CWE-79 is updated, we can see the changelog of CWE-79.json only. In addition, we sometimes want to see the data on GitHub, at least in the case of CVE. In that case, if the `cwe.json` is too big, it can't be opened due to the size or takes a while.

cwe/cwe_test.go
```diff
@@ -0,0 +1,95 @@
...
+			dir, _ := ioutil.TempDir("", "TestUpdate-*")
+			defer func() {
+				_ = os.RemoveAll(dir)
+			}()
+
+			c := NewCWEWithConfig(cweURL, filepath.Join(dir), 0)
+			err := c.Update()
+			switch {
+			case tc.expectedError != "":
+				assert.Contains(t, err.Error(), tc.expectedError, tc.name)
```
knqyf263, line 79: If `err` is nil, it will be panicking. IDE always warns me, so I insert `require.NotNil(t, err)`.

```suggestion
				require.NotNil(t, err)
				assert.Contains(t, err.Error(), tc.expectedError, tc.name)
```

simar7: Fixed.

simar7:
> I also don't think `cwe.xml` is necessary in `vuln-list`. Other sources don't store the original data.
> Also, what if we store CWE-IDs separately such as CWE-79.json? It enables us to track the data easily. If CWE-79 is updated, we can see the changelog of CWE-79.json only. In addition, we sometimes want to see the data on GitHub, at least in the case of CVE. In that case, if the `cwe.json` is too big, it can't be opened due to the size or takes a while.

I agree. I updated it to be that.

cwe/cwe.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type CWEConfig struct {
+	url        string
+	retryTimes int
+	cweDir     string
+}
+
+func NewCWEConfig() CWEConfig {
+	return NewCWEWithConfig("https://cwe.mitre.org/data/xml/cwec_latest.xml.zip", filepath.Join(utils.VulnListDir(), "cwe"), 5)
```
knqyf263, line 25: nit: URL is defined as `const` in other sources.

cwe/cwe.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (c CWEConfig) saveFile(b []byte, fileType string) error {
+	if err := os.MkdirAll(c.cweDir, os.ModePerm); err != nil {
+		return xerrors.Errorf("unable to create cwe directory: %w", err)
+	}
```
knqyf263, line 70: nit: I think we don't need to try creating the directory every time. I mean we can define this process outside of `saveFile` and the iteration of `wc.Weakness.Weakness`.

cwe/cwe_test.go
```diff
@@ -0,0 +1,90 @@
...
+					b, _ := ioutil.ReadFile(tc.inputZipFile)
+					_, _ = io.WriteString(w, string(b))
+				}))
+				cweURL = ts.URL
+				defer func() {
+					ts.Close()
+				}()
+			}
+
+			dir, _ := ioutil.TempDir("", "TestUpdate-*")
```
knqyf263, line 70: nit: We can ignore this error, but it will introduce another error at `c.Update()` when it fails to create the temp dir. In that case, it may obscure the root cause.

cwe/cwe.go
simar7: fixed

cwe/cwe.go
simar7: good point! I've fixed that.

cwe/cwe_test.go
simar7: I think it's fine as is, it would be a failure either way.

knqyf263: Yes, it's no big deal, but I mean it would take a long time to figure out the root cause. For example,

```
a, _ := func1()
b, _ := func2()
c, _ := func3()
err := func4(a, b, c) // <- this might fail due to the error of func1/2/3
assert.NoError(t, err)
```

In this case, as you said, it would be a failure either way at func4(). But it is difficult to know if the error comes from func1, func2, func3, or func4. If func1() creates an instance (in this example, `a`) representing a network socket, func4() may cause a panic of nil pointer when the network connection can't be established and `a` is nil. The error from func1() should describe the detail like `connect to address 127.0.0.1: Connection refused`. When we get a nil pointer panic, we can't know that the root cause is a network problem quickly. If we have assertions for each error, I think it would be helpful for debugging.
In addition, it might cause unexpected behavior. If `ioutil.TempDir` fails and returns an empty string, the files are created under the current directory and the current directory will be removed after this test because it has `defer`.

```
			defer func() {
				_ = os.RemoveAll(dir)
			}()
```

I don't strongly recommend adding the assertion as I said `nit`, but for the reasons above, I basically prefer adding an error check for each err if I don't have a reason to ignore the error.

PR #47, opened 2020-08-27
Fix: Migrate Red Hat Security Data API to OVAL v2

rahul2393:
* Added compare script to print missing CVEs in RedHat and RedHat-oval
* Updated oval/Redhat to parse unpatched vulnerability

### DIFF
--- a/oval/redhat/redhat.go
+++ b/oval/redhat/redhat.go
@@ -9,7 +9,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/spf13/afero"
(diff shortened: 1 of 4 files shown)


oval/redhat/redhat.go
```diff
@@ -21,46 +21,60 @@ import (
 const (
 	ovalDir   = "oval"
 	redhatDir = "redhat"
-	urlFormat = "https://www.redhat.com/security/data/oval/v2/RHEL%s/rhel-%s.oval.xml.bz2"
-	retry     = 5
+
+	classVulnerability = "vulnerability"
+
+	urlFormat                   = "https://www.redhat.com/security/data/oval/v2/RHEL%s/rhel-%s.oval.xml.bz2"
+	urlFormatIncludingUnpatched = "https://www.redhat.com/security/data/oval/v2/beta/RHEL%s/rhel-%s-including-unpatched.oval.xml.bz2"
```
knqyf263, line 28: They seem to already have including-unpatched under RHEL 6,7,8, not `beta/`. Also, we have to figure out the difference between rhel-6-extras-including-unpatched.oval.xml.bz2 and rhel-6-including-unpatched.oval.xml.bz2.

PR #58, opened 2020-11-01
fix(cvrf/suse): handle invalid UTF-8 characters

knqyf263:
Fix the following issue
https://github.com/aquasecurity/vuln-list-update/runs/1337105587?check_suite_focus=true

SUSE recently added invalid UTF-8 characters to their CVRF. I'm sure it is a bug, but `vuln-list-update` should be able to handle it. This PR removes all invalid characters with `strings.ToValidUTF8`.

### DIFF
--- a/cvrf/suse/suse.go
+++ b/cvrf/suse/suse.go
@@ -86,8 +86,7 @@ func (c Config) update(os string, urls []string) error {
 
 		if !utf8.Valid(cvrfXml) {
 			log.Println("invalid UTF-8")
-			s := strings.ToValidUTF8(string(cvrfXml), "")
-			cvrfXml = []byte(s)
+			cvrfXml = []byte(strings.ToValidUTF8(string(cvrfXml), ""))
 		}
 
 		err = xml.Unmarshal(cvrfXml, &cv)
--- a/cvrf/suse/suse.go
+++ b/cvrf/suse/suse.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/aquasecurity/vuln-list-update/utils"
 	"github.com/cheggaaa/pb"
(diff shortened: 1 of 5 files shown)


cvrf/suse/suse.go
```diff
@@ -79,9 +80,16 @@ func (c Config) update(os string, urls []string) error {
...
 		if len(cvrfXml) == 0 {
-			log.Printf("empty CVRF xml: %s", cv.Tracking.ID)
+			log.Println("empty CVRF xml")
 			continue
 		}
+
+		if !utf8.Valid(cvrfXml) {
+			log.Println("invalid UTF-8")
+			s := strings.ToValidUTF8(string(cvrfXml), "")
+			cvrfXml = []byte(s)
```
simar7, line 90: inline
```suggestion
			cvrfXml = []byte(strings.ToValidUTF8(string(cvrfXml), ""))
```

cvrf/suse/suse.go
```diff
@@ -79,9 +80,16 @@ func (c Config) update(os string, urls []string) error {
 	for _, cvrfXml := range cvrfXmls {
 		var cv Cvrf
 		if len(cvrfXml) == 0 {
-			log.Printf("empty CVRF xml: %s", cv.Tracking.ID)
+			log.Println("empty CVRF xml")
 			continue
 		}
+
+		if !utf8.Valid(cvrfXml) {
+			log.Println("invalid UTF-8")
```
simar7, line 88: nit: Should we print the ID for debugging purposes?

knqyf263: I wanted to do that, but we don't have any ID here because it happens before parsing XML.

cvrf/suse/suse.go
knqyf263: Just my rule: No more than two processes in one line for readability
1. string(cvrfXml)
1. strings.ToValidUTF8(b)
1. []byte(s)

But it is not so strong rule, so I follow your suggestion here.

cvrf/suse/suse.go
simar7: I see.

simar7 (approved):
lgtm!

PR #61, opened 2020-12-08
Continue on failure with alerting.

simar7:
This commit helps to continue the build if one or more advisories fail to run.

It also sends a notification into Microsoft Teams channel for us to be alerted.

TODO:
- [x] Add logic to continue after failure.
- [ ] Create a web hook in the appropriate channel
- [ ] Add web hook URL as a secret to this repo

Signed-off-by: Simarpreet Singh <simar@linux.com>

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -181,3 +181,29 @@ jobs:
             targets:
                   - os: "default"
                     uri: "https://github.com/aquasecurity/vuln-list-update/actions?query=is%3Afailure+event%3Aschedule"
+
+    - name: Check for failures and Fail the build
+      if: >
+        steps.nvd.outputs.status != 'success' ||
+        steps.alpine.outputs.status != 'success' ||
+        steps.debian.outputs.status != 'success' ||
+        steps.debian-oval.outputs.status != 'success' ||
+        steps.ubuntu.outputs.status != 'success' ||
+        steps.amazon.outputs.status != 'success' ||
+        steps.oracle-oval.outputs.status != 'success' ||
+        steps.redhat-oval.outputs.status != 'success' ||
+        steps.photon.outputs.status != 'success' ||
+        steps.ghsa.outputs.status != 'success' ||
+        steps.cwe.outputs.status != 'success' ||
+        steps.suse-cvrf.outputs.status != 'success' ||
+        steps.redhat-1996-2002.outputs.status != 'success' ||
+        steps.redhat-2003-2008.outputs.status != 'success' ||
+        steps.redhat-2012.outputs.status != 'success' ||
+        steps.redhat-2013.outputs.status != 'success' ||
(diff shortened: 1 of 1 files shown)


.github/workflows/update.yml
```diff
@@ -31,70 +31,153 @@ jobs:
...
+        steps.redhat-1996-2002.outputs.status != 'success' ||
+        steps.redhat-2003-2008.outputs.status != 'success' ||
+        steps.redhat-2012.outputs.status != 'success' ||
+        steps.redhat-2013.outputs.status != 'success' ||
+        steps.redhat-2014.outputs.status != 'success' ||
+        steps.redhat-2015.outputs.status != 'success' ||
+        steps.redhat-2016.outputs.status != 'success' ||
+        steps.redhat-2017.outputs.status != 'success' ||
+        steps.redhat-2018.outputs.status != 'success' ||
+        steps.redhat-2019-2020.outputs.status != 'success'
```
simar7, line 171: I couldn't find a way to be able to classify and gather the error on a job level but only on a step level. Hence we need these conditions. Any suggestions are welcome if this is unacceptable. 

To me since this logic will rarely change, I think it's fine as is.

knqyf263: Can we know which job fails?

.github/workflows/update.yml
```diff
@@ -31,70 +31,153 @@ jobs:
...
+        summary: 'vuln-list-update failed'
+        title: 'vuln-list-update'
+        text: 'One or more vendor advisory updates have failed. Check link for details.'
+        theme-color: 'Error'
+        potential-action: |
+          - "@type": "OpenUri"
+            name: "View Failures"
+            targets:
+                  - os: "default"
+                    uri: "https://github.com/aquasecurity/vuln-list-update/actions?query=is%3Afailure+event%3Aschedule"
```
knqyf263, line 183: Recently Photon fails every day due to their issue. I'm afraid this notification would be noisy. What do you think? Even if we notify it, we need a separate channel.

simar7: I'm fine with a separate channel.

.github/workflows/update.yml
simar7: After reading the docs, I couldn't find a way.

simar7:
@knqyf263 - I did however manage to fail the build incase an actual failure of running vuln-list-update happens. You can check for an example here: https://github.com/simar7/vuln-list-update/runs/1534495562?check_suite_focus=true

knqyf263:
Hmm. It still doesn't look like we can know which job failed.

simar7:
> Hmm. It still doesn't look like we can know which job failed.

I couldn't find a way to do it. Would that be a deal breaker?

knqyf263:
We have a lot of jobs and I'm sure it bothers us to click and open all jobs every time one of them fails.

PR #69, opened 2021-01-12
feat: clean repository at the end

knqyf263:
If the process fails in the middle, cruft will be left behind.


masahiro331 (approved):

masahiro331:
LGTM!

PR #70, opened 2021-01-12
chore(actions): run steps always

knqyf263:
Even if one of the steps fails, the remaining steps should be continued.
https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#always

image

Only failure steps would be red.


simar7 (approved):
Nice!

Issue #77, opened 2021-03-10
fix(ubuntu): non-existent CVE-2019-5144


knqyf263:
Looks like Ubuntu no longer has CVE-2019-5144.
https://people.canonical.com/~ubuntu-security/cve/2019/CVE-2019-5144

But vuln-list has it.
https://github.com/aquasecurity/vuln-list/blob/37d8ac3154161acd547d306cd536b8024c4cc251/ubuntu/2019/CVE-2019-5144.json

We have to figure out why it happened and address it.

ronaudinho:
[diff is grabbed but not used here](https://github.com/aquasecurity/vuln-list-update/blob/1e28a8e1508b0c54812dea9aff2b00304c5f1221/ubuntu/ubuntu.go#L70) and walk is based on existing files in `ubuntu-cve-tracker`, meaning if file is deleted there but exists in `vuln-list`, we will miss updating it.
i think we can either:
1. take note of changed files in `ubuntu-cve-tracker` and only updates the corresponding files in `vuln-list`
2. delete target dir in `vuln-list` and start fresh every time

option 1 is perhaps more error-prone and it needs to be done on `git pull`, which is probably not feasible in CI (I am not sure how long we can store stuffs in github actions cache), but has less impact radius if failure happens
option 2 is a lot simpler and can be done on `git clone`, but should take rather longer, especially given the number of files being processed. it also has bigger impact radius if it fails

wdyt @knqyf263

knqyf263:
Let's take option 2. If it fails, we don't commit the status. `vuln-list` will keep the previous commit as is.

PR #79, opened 2021-03-12
ubuntu: remove dir if clone/pull fails halfway

### LINKED ISSUE #76: fix(ubuntu): status disparency (CVE-2019-12379)
The status is ignored in the case of `linux-azure-edge`.
https://people.canonical.com/~ubuntu-security/cve/2019/CVE-2019-12379

Ubuntu CVE Tracker also says like that.
https://git.launchpad.net/ubuntu-cve-tracker/tree/ignored/CVE-2019-12379

However, `vuln-list` says `needed`.
https://github.com/aquasecurity/vuln-list/blob/37d8ac3154161acd547d306cd536b8024c4cc251/ubuntu/2019/CVE-2019-12379.json#L207-L210

CVE-2019-12378 is the same situation.

ronaudinho:
resolves #76 and #77. failing fetch into existing git dir causes the next clone/pull attempt to fail, this adds removing said dir before retrying clone/pull.

### DIFF
--- a/ubuntu/ubuntu.go
+++ b/ubuntu/ubuntu.go
@@ -78,15 +78,14 @@ func Update() error {
 	dir := filepath.Join(utils.CacheDir(), cveTrackerDir)
 	for _, url := range repoURLs {
 		_, err = gc.CloneOrPull(url, dir, "master", false)
-		if err != nil {
-			log.Printf("failed to clone or pull: %s: %v", url, err)
-			log.Printf("removing %s directory", cveTrackerDir)
-			if err := os.RemoveAll(dir); err != nil {
-				return xerrors.Errorf("failed to remove %s directory: %w", cveTrackerDir, err)
-			}
-			continue
+		if err == nil {
+			break
 		}
-		break
+		log.Printf("failed to clone or pull: %s: %v", url, err)
+		log.Printf("removing %s directory", cveTrackerDir)
+		if err := os.RemoveAll(dir); err != nil {
+			return xerrors.Errorf("failed to remove %s directory: %w", cveTrackerDir, err)
+		}
 	}
 	if err != nil {
 		return xerrors.Errorf("failed to clone or pull: %w", err)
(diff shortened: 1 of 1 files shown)


ubuntu/ubuntu.go
```diff
@@ -80,6 +80,10 @@ func Update() error {
 		_, err = gc.CloneOrPull(url, dir, "master", false)
 		if err != nil {
```
ronaudinho, line 81:
```
...
  if err == nil {
    break
  }
  log.Printf("failed to clone or pull: %s: %v", url, err)
  log.Printf("removing %s directory", cveTrackerDir)
  if err := os.RemoveAll(dir); err != nil {
    return xerrors.Errorf("failed to remove %s directory: %w", cveTrackerDir, err)
  }
}
```
looks cleaner to me here

knqyf263: OK, let's change it to what you suggest.

ubuntu/ubuntu.go
```diff
@@ -78,11 +78,14 @@ func Update() error {
...
 		_, err = gc.CloneOrPull(url, dir, "master", false)
-		if err != nil {
-			log.Printf("failed to clone or pull: %s: %v", url, err)
-			continue
+		if err == nil {
+			break
+		}
+		log.Printf("failed to clone or pull: %s: %v", url, err)
+		log.Printf("removing %s directory", cveTrackerDir)
+		if err := os.RemoveAll(dir); err != nil {
```
knqyf263, line 86: Even if the directory doesn't exist, doesn't it return an error? I'm concerned that the first try fails immediately and it didn't create a directory. Then, `os.RemoveAll` fails and the second try will not be performed.

ronaudinho: actually that makes sense, I'll update the PR.

ronaudinho:
```
// it encounters. If the path does not exist, RemoveAll
// returns nil (no error).
```
actually from [os.RemoveAll docs](https://github.com/golang/go/blob/d0d38f0f707e69965a5f5a637fa568c646899d39/src/os/path.go#L63-L64), so if dir does not exist, this will not matter.

knqyf263: Thank you for confirming.

knqyf263 (approved):

Issue #57, opened 2020-10-18
Support safety-db


knqyf263:
Currently, `trivy-db` downloads `pyupio/safety-db` directly.
https://github.com/aquasecurity/trivy-db/blob/master/Makefile#L73

But if `safety-db` changes the format, it causes the DB build to fail. We should parse `safety-db` and commit it to `vuln-list` once.

The directory structure should be like the following.

- python
  - safety-db
    - django
      - pyup.io-25694.json
      - pyup.io-25695.json
    - django-access-tokens

knqyf263:
@rahul2393 Could you work on it?

rahul2393:
Ack

PR #84, opened 2021-04-27
Add python safety db

rahul2393:
[Issue](https://github.com/aquasecurity/vuln-list-update/issues/57)


knqyf263:
We'll migrate to OSV.

PR #92, opened 2021-06-21
Added Debian salsa also into Tracker

tonaim:
1. Capture security Advisory DSA and DLA
2. Relation between security advisory and CVE
3. Support EOL version in trivy. Currently Trivy supports CVEs which have fixed versions. But we should support open CVEs also.
4. Capture vendor severity
5. Capture will not fix CVEs

### DIFF
--- a/debian/salsa/debian.go
+++ b/debian/salsa/debian.go
@@ -1,0 +1,528 @@
+package salsa
+
+import (
+	"bufio"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/PuerkitoBio/goquery"
+
+	"golang.org/x/xerrors"
+
+	"github.com/cheggaaa/pb"
+	"github.com/pkg/errors"
+
+	"github.com/aquasecurity/vuln-list-update/utils"
+)
+
+const (
+	cloneCmd           = "clone --depth 1 https://salsa.debian.org/security-tracker-team/security-tracker.git"
+	debianDir          = "debian-salsa"
+	coveredReleasesURL = "http://security-tracker.debian.org/tracker/data/releases"
+	releasesURL        = "http://www.debian.org/releases/"
+)
+
+var releaseRegexp = regexp.MustCompile(`Debian (?:GNU/Linux )?(\d+)`)
+
+type dsa struct {
+	name        string
+	date        time.Time
+	description string
+	cves        []string
+	packages    []pkg
+}
+
+type pkg struct {
+	release, name, version, severity, statement string
+	willNotFix                                  bool
+	classification                              int64
+	severityClassification                      string
+}
+
+type cve struct {
+	name     string
+	reserved bool
+	notForUs bool
+	packages []pkg
+	notes    []string
+}
+
+var (
+	dsaLine     = regexp.MustCompile(`^\[(\d\d \w{3} \d{4})\] (D(?:S|L)A-\d+(?:-\d+)?)\s*(.*)`)
+	dsaCVEsLine = regexp.MustCompile(`(CVE-\d+-\d+)`)
+	dsaPkgLine  = regexp.MustCompile(`^\s*(?:\[(\w+)\]\s*)?- ([^\s]+) ([^\s]+)$`)
+)
+
+var (
+	cveLine      = regexp.MustCompile(`^(CVE-\d+-[0-9X]+)`)
+	notForUsLine = regexp.MustCompile(`^\s*NOT-FOR-US`)
+	noteLine     = regexp.MustCompile(`^\s*NOTE:\s*(.+)$`)
+	reservedLine = regexp.MustCompile(`^\s*RESERVED`)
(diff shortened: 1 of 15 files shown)


debian/salsa/debian.go
```diff
@@ -0,0 +1,537 @@
...
+					}
+				}
+			}
+		}
+		dsaFile.Close() // nolint: errcheck
+	}
+	return nil
+}
+
+func (ctx *DebianSalsa) parseCVEs() error {
```
knqyf263, line 317: Why do we need to parse CVE? For EOL version? Are DLA/DSA not enough? Can we obtain metadata such as severity from security tracker?
https://security-tracker.debian.org/tracker/data/json

And this method is not DRY-ed. Please refactor it.

knqyf263 (commented):
Please write tests

debian/salsa/debian.go
tonaim: yes we would like to parse CVEs also, because we can find open CVEs for EOL versions here. Example CVE-2020-15395 is not fixed in jessie release and we can get that information here. 
image

debian/salsa/debian_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/xerrors"
+)
+
+func TestDebianSalsa_getReleases(t *testing.T) {
```
knqyf263, line 16: Please write tests only for Update().

debian/salsa/debian_test.go
```diff
@@ -0,0 +1,157 @@
+package salsa
```
knqyf263, line 1:
```suggestion
package salsa_test
```

debian/salsa/debian_test.go
```diff
@@ -0,0 +1,157 @@
...
+				t.Errorf("cveToDSA = %v, wantCVEtoDSA %v", debianCtx.cveToDSA, tt.wantCVEtoDSA)
+			}
+			if err := debianCtx.parseCVEs(); (err != nil) != tt.wantErr {
+				t.Errorf("Update() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if err := debianCtx.writePackages(); (err != nil) != tt.wantErr {
+				t.Errorf("writePackages() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			dir := filepath.Join(debianCtx.VulnListDir, "debian-salsa")
+			assert.NoError(t, err, "failed to create temp dir")
```
knqyf263, line 130: Please refer to other tests.

debian/salsa/debian_test.go
```diff
@@ -0,0 +1,157 @@
...
+		"CVE-2021-33477": {dla2681}, "CVE-2002-0653": {dsa135}, "CVE-2014-3466": {dla0001}}
+
+	tests := []struct {
+		name         string
+		fields       fields
+		wantErr      bool
+		wantOss      map[string]string
+		wantCVEtoDSA map[string][]dsa
+	}{
+		{name: "All functions", fields: releaseFields, wantErr: false, wantOss: ossData, wantCVEtoDSA: wantCVEtoDSAData},
```
knqyf263, line 99: Keep a single case small and should add more cases.

debian/salsa/debian_test.go
```diff
@@ -0,0 +1,157 @@
...
+					return xerrors.Errorf("walk error: %w", err)
+				}
+				if info.IsDir() {
+					return nil
+				}
+				paths := strings.Split(path, string(os.PathSeparator))
+				p := filepath.Join(paths[len(paths)-2:]...)
+				golden := filepath.Join("testdata", "golden", "debian-salsa", p+".golden")
+
+				want, err := ioutil.ReadFile(golden)
```
knqyf263, line 143: ioutil is deprecated

debian/salsa/debian_test.go
```diff
@@ -0,0 +1,157 @@
...
+func TestDebianSalsa_getReleases(t *testing.T) {
+	type fields struct {
+		VulnListDir    string
+		oss            map[string]string
+		cveToDSA       map[string][]dsa
+		PackageData    map[string]map[string]CVERelease
+		cloneDirectory string
+	}
+	releaseFields := fields{
+		VulnListDir:    "testdata/got",
```
knqyf263, line 25: Should use t.TempDir()

debian/salsa/debian.go
```diff
@@ -0,0 +1,469 @@
...
+	return &DebianSalsa{
+		VulnListDir:    utils.VulnListDir(),
+		cveToDSA:       make(map[string][]dsa),
+		PackageData:    make(map[string]map[string]CVERelease),
+		cloneDirectory: salsaDebianCloneDir(),
+	}
+}
+
+func (ctx DebianSalsa) Update() error {
+	gc := git.Config{}
```
knqyf263, line 135: Doesn't tarball work?
https://salsa.debian.org/security-tracker-team/security-tracker/-/archive/master/security-tracker-master.tar.gz

knqyf263 (changes requested):

knqyf263:
In favor of #95

PR #93, opened 2021-07-08
feat(alpine): add support of unfixed vulnerabilities

tonaim:
Now, we can identify the packages, which can have vulnerabilities based on NVD data


alpine/alpine-unfixed/alpine.go
```diff
@@ -0,0 +1,202 @@
...
+	if len(errs) > 0 {
+		return nil, xerrors.Errorf("HTTP error. url: %s, err: %w", url, errs[0])
+	}
+	if resp.StatusCode != 200 {
+		return nil, xerrors.Errorf("HTTP error. status code: %d, url: %s", resp.StatusCode, url)
+	}
+	return body, nil
+}
+
+func cveHTMLParser(baseUrl, vulnerability string, baseData map[string]string) error {
```
knqyf263, line 173: Why do we need to parse CVE pages?

alpine/alpine-unfixed/alpine.go
```diff
@@ -0,0 +1,202 @@
...
+	if err := u.appFs.MkdirAll(dir, 0755); err != nil {
+		return xerrors.Errorf("Failed creating dir %s: %w", dir, err)
+	}
+
+	log.Println("Fetching branch data...")
+	activeBranches, err := getActiveReleases(u.baseURL)
+	if err != nil {
+		return xerrors.Errorf("Failed getting active branches: %w", err)
+	}
+	activeBranches = append(activeBranches, "3.10-main")
```
knqyf263, line 60: Let's start with hard-coded versions. we may be able to contribute to their source. I think `show_index` should support JSON as well.
https://gitlab.alpinelinux.org/kaniini/secfixes-tracker/-/blob/master/secfixes_tracker/views.py#L9-12

alpine/alpine-unfixed/alpine.go
```diff
@@ -0,0 +1,202 @@
...
+	req.Header.Add("Accept", acceptHeader)
+	resp, body, errs := req.Type("text").EndBytes()
+	if len(errs) > 0 {
+		return nil, xerrors.Errorf("HTTP error. url: %s, err: %w", url, errs[0])
+	}
+	if resp.StatusCode != 200 {
+		return nil, xerrors.Errorf("HTTP error. status code: %d, url: %s", resp.StatusCode, url)
+	}
+	return body, nil
+}
```
knqyf263, line 171: Expand this function.
https://github.com/aquasecurity/vuln-list-update/blob/main/utils/utils.go#L123-L136

alpine/alpine-unfixed/alpine_test.go
```diff
@@ -0,0 +1,131 @@
...
+					if expVulns, ok := expectedJson.UnfixVersion[version]; !ok {
+						assert.True(t, ok)
+					} else {
+						sort.Strings(actualVuls)
+						sort.Strings(expVulns)
+						assert.True(t, reflect.DeepEqual(actualVuls, expVulns))
+					}
+				}
+				assert.True(t, expectedJson.DistroVersion == actualJson.DistroVersion)
+				require.True(t, reflect.DeepEqual(expectedJson, actualJson), "")
```
knqyf263, line 123: Is there any reason that we cannot use assert.Equal?
```suggestion
				assert.Equal(t, expectedJson, actualJson)
```

alpine/alpine-unfixed/alpine.go
tonaim: no it does not support JSON
https://gitlab.alpinelinux.org/kaniini/secfixes-tracker/-/blob/master/secfixes_tracker/views.py#L9-12
has only 
`@accept('text/html')`

knqyf263: I know it is not supported now. I meant we can contribute to secfixes_tracker so that it can support JSON API in `show_index`.

alpine/alpine-unfixed/alpine.go
tonaim: we have to parse CVE html page. Consider below example.
1. Use https://security.alpinelinux.org/branch/edge-community to all vulnerabilities and packages in edge-community version in json format.
2. Take a CVE(https://security.alpinelinux.org/vuln/CVE-2019-12212) from the above response and check in JSON.

data will be like below screenshot
image

But if you see same in html response 
image

In JSON you cannot distinguish which state object belongs to which release. But in HTML response, you can clearly see that.

alpine/alpine-unfixed/alpine_test.go
tonaim: Fixed it

alpine/alpine-unfixed/alpine.go
tonaim: Fixed it

alpine/alpine-unfixed/alpine.go
knqyf263: This endpoint returns "Potentially vulnerable packages in edge-community". So, you don't need to check the HTML. You already know "freeimage" is potentially vulnerable to "CVE-2019-12212".
https://security.alpinelinux.org/branch/edge-community

Also, you don't need to worry about 3.14-community. CVE-2019-12212 is listed as below.
https://security.alpinelinux.org/branch/3.14-community

tonaim: Yes, but we need to get the package version also.

knqyf263: Why? It is not a fixed version.

tonaim: yes it not fixed version. But we want to consider it as affected to version. Means until this version package is vulnerable.

utils/fs.go
```diff
@@ -31,3 +32,12 @@ func WriteJSON(fs afero.Fs, dir, fileName string, data interface{}) error {
...
 }
+
+func JSONMarshal(t interface{}) ([]byte, error) {
+	buffer := &bytes.Buffer{}
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+	err := encoder.Encode(t)
+	return buffer.Bytes(), err
+}
```
knqyf263, line 43: Why do we need this?

alpine/alpine-unfixed/types.go
```diff
@@ -0,0 +1,21 @@
...
+	PackageName    string `json:"packageName,omitempty"`
+	PackageVersion string `json:"packageVersion,omitempty"`
+}
+
+type CPEMatch struct {
+	MinVersion    string `json:"minimumVersion"`
+	MinVersionOps string `json:"minimumVersionOp"`
+	MaxVersion    string `json:"maximumVersion"`
+	MaxVersionOps string `json:"maximumVersionOp"`
+}
```
knqyf263, line 21: I don't think it is necessary.

alpine/alpine-unfixed/alpine.go
```diff
@@ -0,0 +1,129 @@
...
+	defer os.RemoveAll(u.fileDownloadPath) // nolint: errcheck
+
+	err = u.ExtractTarGz(dir)
+	if err != nil {
+		return err
+	}
+	return
+}
+
+func (u Updater) ExtractTarGz(saveDir string) error {
```
knqyf263, line 67: Could you rewrite this function by using https://github.com/aquasecurity/vuln-list-update/pull/105?

knqyf263 (changes requested):

alpine/alpine-unfixed/types.go
tonaim: So the version that is present in state is affected version. But we can get range from CPE. Example:
image
By looking at CPE, I can tell that beep package is vulnerable **from** 0 **to** 1.3.4 version

utils/fs.go
tonaim: This is to parse CPE. We have '<'  and '>' symbols in CPEMatch

alpine/alpine-unfixed/types.go
knqyf263: As I explained earlier, it is the same as NVD.

knqyf263 (approved):

PR #95, opened 2021-07-20
feat(debian): replace JSON API with salsa repo

knqyf263:
# Overview
Replace https://github.com/aquasecurity/vuln-list-update/pull/92

# TODO
- [x] data/DLA/list
- [x] data/DSA/list
- [x] data/CVE/list
- [x] static/distributions.json
- [x] https://ftp.debian.org/debian/dists/%s/%s/source/Sources.gz
- [x] https://security.debian.org/debian-security/dists/%s/updates/%s/source/Sources.gz


debian/tracker/annotation.go
```diff
@@ -0,0 +1,256 @@
...
+	for _, p := range d.parsers {
+		match := p.Match(line)
+		if len(match) != 0 {
+			once.Do(func() {
+				ann = &Annotation{
+					Original: strings.TrimSpace(line),
+					Line:     lineno,
+				}
+			})
+			p.Apply(match, ann)
```
tonaim, line 94: we can break the loop if we find the match.

knqyf263: I'm concerned that a single line contains multiple annotations, so I didn't break the loop. Are you sure multiple annotations are not allowed?

ankk13 (approved):

PR #83, opened 2021-04-23
Added target go-vulndb

rahul2393:
[Issue](https://github.com/aquasecurity/vuln-list-update/issues/82)

PR #96, opened 2021-08-11
feat(python): add pypa support

AndreyLevchenko:


pypa/pypa.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+type options struct {
+	url   string
+	dir   string
+	retry int
+}
+type option func(*options)
+
+type Pypa struct {
```
knqyf263, line 32:
```suggestion
type PyPA struct {
```

According to their website, PyPA is correct.
https://www.pypa.io/en/latest/

pypa/pypa.go
```diff
@@ -0,0 +1,125 @@
...
+	pypaDir            = "pypa"
+	securityTrackerURL = "https://github.com/pypa/advisory-db/archive/refs/heads/main.zip"
+	retry              = 3
+	yamlExt            = ".yaml"
+)
+
+type options struct {
+	url   string
+	dir   string
+	retry int
```
knqyf263, line 28: `retry` seems to be unused.

pypa/pypa.go
```diff
@@ -0,0 +1,125 @@
...
+	yamlFiles, err := getYamlFiles(vulnDir)
+
+	if err != nil {
+		return xerrors.Errorf("failed to find vulnerability files in the directory %s: %w", vulnDir, err)
+	}
+
+	bar := pb.StartNew(len(yamlFiles))
+
+	for _, file := range yamlFiles {
+		data, err := ioutil.ReadFile(file)
```
knqyf263, line 85: ioutil is deprecated
```suggestion
		data, err := os.ReadFile(file)
```

pypa/pypa_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+			defer ts.Close()
+
+			// Intentionally close to induce network errors
+			if tt.inputArchive == "" {
+				ts.Close()
+			}
+
+			dir := t.TempDir()
+			c := NewPypa(WithURL(ts.URL+"/"+tt.inputArchive), WithDir(filepath.Join(dir)), WithRetry(0))
```
knqyf263, line 49: You have to pass `afero.NewMemMapFs()` and need to use the filesystem below instead of `os.ReadDir` and `os.ReadFile`.

pypa/pypa_test.go
```diff
@@ -0,0 +1,78 @@
+package pypa
```
knqyf263, line 1:
```suggestion
package pypa_test
```

pypa/pypa_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Update(t *testing.T) {
+	tests := []struct {
+		name          string
+		inputArchive  string
+		expectedFiles []string
```
knqyf263, line 18:
```suggestion
		wantFiles []string
```

pypa/pypa_test.go
```diff
@@ -0,0 +1,78 @@
...
+			dir := t.TempDir()
+			c := NewPypa(WithURL(ts.URL+"/"+tt.inputArchive), WithDir(filepath.Join(dir)), WithRetry(0))
+			err := c.Update()
+			if tt.wantErr != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.wantErr)
+				return
+			} else {
+				require.NoError(t, err)
+			}
```
knqyf263, line 57:
```suggestion
			}
			
			require.NoError(t, err)
```

main.go
```diff
@@ -156,6 +157,12 @@ func run() error {
 			return xerrors.Errorf("error in Photon update: %w", err)
 		}
 		commitMsg = "Photon Security Advisories"
+	case "pypa":
+		p := pypa.NewPypa()
+		if err := p.Update(); err != nil {
+			return xerrors.Errorf("error in Pypa update: %w", err)
```
knqyf263, line 163:
```suggestion
			return xerrors.Errorf("error in PyPA update: %w", err)
```

main.go
```diff
@@ -156,6 +157,12 @@ func run() error {
...
+		}
+		commitMsg = "Pypa Security Advisories"
```
knqyf263, line 165:
```suggestion
		commitMsg = "PyPA Security Advisories"
```

pypa/pypa.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (pypa *Pypa) Update() error {
+	dir, err := utils.DownloadToTempDir(context.Background(), pypa.opts.url)
+
+	if err != nil {
+		return xerrors.Errorf("failed to download %s: %w", pypa.opts.url, err)
+	}
+
+	vulnDir := filepath.Join(dir, "advisory-db-main/vulns")
```
knqyf263, line 74:
```suggestion
	vulnDir := filepath.Join(dir, "advisory-db-main", "vulns")
```

pypa/pypa.go
```diff
@@ -0,0 +1,125 @@
...
+
+		osv := &types.Osv{}
+
+		err = yaml.Unmarshal(data, osv)
+
+		if err != nil {
+			return xerrors.Errorf("unable to parse yaml %s: %w", file, err)
+		}
+
+		if err := utils.WriteJSON(pypa.AppFs, pypa.opts.dir, fmt.Sprintf("%s.json", osv.Id), osv); err != nil {
```
knqyf263, line 99: We should keep the directory structure as much as possible. It means we will have package dirs.
https://github.com/pypa/advisory-db/tree/main/vulns

knqyf263 (changes requested):

pypa/pypa.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/aquasecurity/vuln-list-update/utils"
+	"github.com/cheggaaa/pb"
+	"github.com/spf13/afero"
+	"golang.org/x/xerrors"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	pypaDir            = "pypa"
+	securityTrackerURL = "https://github.com/pypa/advisory-db/archive/refs/heads/main.zip"
```
patryk4815, line 20: why not using https://osv-vulnerabilities.storage.googleapis.com/PyPI/all.zip ?
https://github.com/google/osv/blob/master/README.md#current-data-sources

knqyf263: @AndreyLevchenko Could you check if we can use the above URL?

patryk4815:
I think we should change this PR and https://github.com/aquasecurity/vuln-list-update/pull/83 
Into one implementation for handling `osv` as datasource (like `ghsa` target):
https://github.com/google/osv/blob/master/README.md#current-data-sources
They support multiple source (python/go/etc) with same schema.

knqyf263:
@patryk4815 It sounds like a nice idea. Go data source in OSV used to have an issue (https://github.com/golang/vulndb/issues/1) and we needed to parse the repository ourselves. But after having a quick look, I found they might change API. Let me look into OSV carefully. If we can implement `osv`, it would be better. Thanks.

patryk4815:
@knqyf263 Maybe we need wait for "purl" support - https://github.com/google/osv/issues/64

pypa/pypa.go
AndreyLevchenko: @knqyf263 I think I can use the url above, but I need to understand benefits better. For now it looks pretty same to me.

knqyf263: The benefit is we can have a single implementation for some languages (Python, Go and Rust).

knqyf263:
@westonsteimel Thanks! It looks like PyPA Advisory DB works well for us. We have a problem with Go.

knqyf263:
In favor of https://github.com/aquasecurity/vuln-list-update/pull/113

PR #102, opened 2021-09-04
chore(ci): parallelize update workflow

### LINKED ISSUE #101: refactor : Update vuln-list repo
The last workflow took 1h 27m 44s to pass and the longest step took 19m 28s.

What would you think of splitting this workflow in two step.
The first would build the vuln-list-update executable and upload it with actions/upload-artifact
The second would be a matrix of the differents targets (with their appropriate names) which would download the artifact (actions/download-artifact) and then execute `./vuln-list-update -target ...`

The advantage of matrix is of course parallelization of the tasks.
What I don't know yet to create the right PR is what needs to be know by each step (GITHUB_TOKEN, GITHUB_ENV...)

mmorel-35:
Hi @knqyf263, this is an attemps to shorten update workflow duration.

What do you think ?

Fixes #101 

NB: squash commits instead of merge

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -5,8 +5,25 @@ on:
   workflow_dispatch:
 
 jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: Compile vuln-list-update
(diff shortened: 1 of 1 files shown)


knqyf263 (commented):
vuln-list-update pushes commits to vuln-list. If we do these updates in parallel, they will be conflicted.

mmorel-35:
I see, so what it needs is a kind of map reduce system. The retrieving of the content is parallel but the persistence happens only once.

knqyf263:
Exactly.

mmorel-35:
Ok, so I drop this PR. It can be reopen when the projet is map reduce compatible.

PR #108, opened 2021-09-27
bug(debian-severity): add missing severities

ankk13:
Updated severity regex to consider text like  `(<severity>; bug #<no>)`


knqyf263 (changes requested):
When you fix a bug, you should add a new test.

Issue #111, opened 2021-11-11
Debian OVAL update error: invalid CVE-ID format


knqyf263:
## Overview
The following error is blocking the Debian OVAL update.

```
2021/11/11 12:15:25 Debian OVAL update error: invalid CVE-ID format: DSA-3475-1 postgresql-9.1
```

https://github.com/aquasecurity/vuln-list-update/actions/runs/1448664152

knqyf263:
@rahul2393 Could you take a look?

PR #112, opened 2021-11-14
feat(debian-oval): removing debian oval from vuln-list-update

rahul2393:
Fixes: #111


knqyf263:
Debian SALSA should have the same advisories as OVAL.
https://github.com/aquasecurity/vuln-list-update/pull/95

knqyf263 (approved):

PR #113, opened 2021-11-17
feat(osv) : add osv support

DmitriyLewen:
## Description
Support OSV format
https://ossf.github.io/osv-schema/#python-vulnerability

## Supported languages
- Go
- Python
- Rust


osv/types.go
```diff
@@ -0,0 +1,48 @@
...
+// type OsvDatabase struct {
+// }
+
+type OsvReference struct {
+	Type string `json:"type,omitempty"`
+	Url  string `json:"url,omitempty"`
+}
+
+//collection based on https://ossf.github.io/osv-schema/
+type OsvJson struct {
```
AndreyLevchenko, line 37: `Json` postfix isn't very good choice as structure can be used some way which is not related to json serialization

osv/osv.go
```diff
@@ -0,0 +1,159 @@
...
+	url           string
+	dir           string
+	ecosystemDirs map[string]string
+}
+
+type option func(*options)
+
+type Osv struct {
+	opts  *options
+	AppFs afero.Fs
```
AndreyLevchenko, line 38: `AppFs` is not required here

osv/osv.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func WithEcosystem(ecosystemDir map[string]string) option {
+	return func(opts *options) {
+		opts.ecosystemDirs = ecosystemDir
+	}
+}
+
+func NewOsv(opts ...option) Osv {
+	ecosystemDirs := make(map[string]string)
```
AndreyLevchenko, line 60: this should be used only if `WithEcosystem()` hasn't been called.

osv/osv_test.go
```diff
@@ -0,0 +1,109 @@
...
+			for _, ecosystem := range tt.ecosystem {
+				testEcosystemDir[ecosystem.name] = ecosystem.dir
+			}
+
+			c := osv.NewOsv(osv.WithURL(testUrl), osv.WithDir(testDir), osv.WithEcosystem(testEcosystemDir))
+			err := c.Update()
+			if tt.wantErr != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.wantErr)
+				return
```
AndreyLevchenko, line 92: I would better use `else`

AndreyLevchenko (changes requested):

osv/osv_test.go
DmitriyLewen: done

osv/osv.go
DmitriyLewen: added check

osv/osv.go
DmitriyLewen: removed

osv/types.go
DmitriyLewen: changed

osv/osv.go
AndreyLevchenko: Just call all `options` and then check if `ecosystemDirs` is still empty. If it is then assign defaults

DmitriyLewen: changed

osv/osv.go
```diff
@@ -0,0 +1,165 @@
...
+		o.dir = filepath.Join(utils.VulnListDir(), osvDir)
+	}
+
+	return Osv{
+		opts: o,
+	}
+}
+
+func (osv *Osv) Update() error {
+	allfiles, err := osv.getAllFiles()
```
knqyf263, line 87: I prefer iterating the ecosystem here and downloading files inside each iteration.

```
for ecoSystem, ecoSystemDir := range osv.opts.ecosystemDirs {
	log.Printf("Updating OSV %s advisories", ecoSystem)
	downloadFiles(ecoSystem, ecoSystemDir)

	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
		f, err := os.Open(path)
		json.NewDecoder(f).Decode()...
	}
}
```

osv/osv.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return xerrors.Errorf("failed to get files: %w", err)
+	}
+
+	bar := pb.StartNew(osv.getAmountFiles(allfiles))
```
knqyf263, line 92: I think it is fast enough. No need to show a progress bar.

osv/types.go
```diff
@@ -0,0 +1,50 @@
+package osv
+
+type OsvAffected struct {
+	Package  *OsvPackage `json:"package,omitempty"`
+	Ranges   []OsvRange  `json:"ranges,omitempty"`
+	Versions []string    `json:"versions,omitempty"`
+
+	// Ecosystem *OsvEcosystem `json:"ecosystem_specific,omitempty"` //The meaning of the values within the object is entirely defined by the ecosystem
```
knqyf263, line 8: You can use `interface{}`

```suggestion
	EcosystemSpecific interface{} `json:"ecosystem_specific,omitempty"`
```

osv/types.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Database  *OsvDatabase  `json:"database_specific,omitempty"` //The meaning of the values within the object is entirely defined by the database
```
knqyf263, line 10: ditto

osv/types.go
DmitriyLewen: Done

osv/types.go
DmitriyLewen: done

osv/osv.go
DmitriyLewen: changed

main.go
```diff
@@ -193,6 +194,12 @@ func run() error {
 			return xerrors.Errorf("AlmaLinux update error: %w", err)
 		}
 		commitMsg = "AlmaLinux Security Advisory"
+	case "osv":
+		p := osv.NewOsv()
+		if err := p.Update(); err != nil {
+			return xerrors.Errorf("error in Osv update: %w", err)
+		}
+		commitMsg = "Osv Security Advisories"
```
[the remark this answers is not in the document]
DmitriyLewen: Done

osv/osv.go
DmitriyLewen: Removed

knqyf263 (approved):

knqyf263:
Thanks

PR #128, opened 2022-01-14
feat(ghsa): store CVSS score/vector for Github Security Advisory

afdesk:
GitHub may have a different CVSS score from NVD, so we have to store CVSS score and Vector values.

this is the first part for [aquasecurity/trivy-db#174](https://github.com/aquasecurity/trivy-db/issues/174)


knqyf263 (approved):

knqyf263:
Thanks

PR #134, opened 2022-02-03
feat(redhat-oval2): capture state of vulnerability

tonaim:
capture the state of vulnerability from redhat oval2 definations


knqyf263 (changes requested):
Can you add tests?

knqyf263 (changes requested):
We can update the existing definitions instead of adding a new definition.

PR #148, opened 2022-04-15
feat(alpine): added parser for alpine eol dates

DmitriyLewen:
## Description
Alpine has [list of EOL dates](https://alpinelinux.org/releases).

Added parser from this url.

### DIFF
--- a/alpine/eof.go
+++ b/alpine/eof.go
@@ -1,0 +1,57 @@
+package alpine
+
+import (
+	"bytes"
+	"github.com/PuerkitoBio/goquery"
+	"github.com/aquasecurity/vuln-list-update/utils"
+	"golang.org/x/xerrors"
+	"log"
+	"strings"
+	"time"
+)
+
+const (
+	eolUrl       = "https://alpinelinux.org/releases/"
+	retryRequest = 1
+)
+
+func getEOFDates(eolUrl string, retry int) (map[string]time.Time, error) {
+	eolDates := make(map[string]time.Time)
+
+	log.Println("Fetching Alpine end of life dates...")
+	b, err := utils.FetchURL(eolUrl, "", retry)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get eol list: %w", err)
+	}
+
+	doc, err := goquery.NewDocumentFromReader(bytes.NewReader(b))
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read eol list: %w", err)
+	}
+
+	doc.Find("tbody tr").Each(func(_ int, tr *goquery.Selection) {
+		var version string
+		var date time.Time
+
+		tr.Find("td").Each(func(ix int, td *goquery.Selection) {
+			switch ix {
+			case 0:
+				version = strings.TrimPrefix(td.Text(), "v") // remove 'v' prefix
+			case 4:
+				d, err := time.Parse("2006-01-02", td.Text())
+				if err != nil {
+					return
+				}
+
+				// Move time to end of day
+				date = d.Add(time.Hour*23 + time.Minute*59 + time.Second*59)
+			}
+		})
+		eolDates[version] = date
+	})
+
+	// edge version doesn't have EOL. Add max date.
+	eolDates["edge"] = time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC)
+
+	return eolDates, nil
+}
--- a/alpine/eof_test.go
+++ b/alpine/eof_test.go
@@ -1,0 +1,41 @@
+package alpine
+
+import (
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func Test_GetAllEOFDates(t *testing.T) {
+	tests := []struct {
+		name     string
+		filepath string
+		want     map[string]time.Time
+	}{
+		{
+			name:     "happy path",
+			filepath: "testdata/eol.html",
+			want: map[string]time.Time{
+				"3.15": time.Date(2023, 11, 01, 23, 59, 59, 0, time.UTC),
+				"3.14": time.Date(2023, 05, 01, 23, 59, 59, 0, time.UTC),
+				"edge": time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC),
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				http.ServeFile(w, r, test.filepath)
+			}))
+			defer ts.Close()
+
+			dates, _ := getEOFDates(ts.URL, 1)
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -221,6 +222,13 @@ func run() error {
 			return xerrors.Errorf("CBL-Mariner Vulnerability Data update error: %w", err)
 		}
 		commitMsg = "CBL-Mariner Vulnerability Data"
+	// EOL dates
+	case "eol-alpine":
+		src := eolalpine.NewConfig()
+		if err := src.Update(); err != nil {
+			return xerrors.Errorf("Alpine end of life dates update error: %w", err)
+		}
+		commitMsg = "CBL-Mariner Vulnerability Data"
```
knqyf263, line 231: It is not CBL-Mariner

eol/alpine/eof_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func Test_GetAllEOFDates(t *testing.T) {
+	tests := []struct {
+		name     string
+		filepath string
+		want     map[string]time.Time
+	}{
+		{
+			name:     "happy path",
```
knqyf263, line 18: Can we add sad path?

eol/alpine/eof.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (config Config) save(dates map[string]time.Time) error {
+	dir := filepath.Join(config.vulnListDir, eolAlpineFolder)
+	if err := os.RemoveAll(dir); err != nil {
+		return xerrors.Errorf("unable to remove github security advisory directory: %w", err)
+	}
+	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+		return xerrors.Errorf("failed to mkdir: %w", err)
+	}
```
knqyf263, line 102: WriteJSON also creates a dir.
https://github.com/aquasecurity/vuln-list-update/blob/357afc54d8e25063171062c9054e5cc210def515/utils/fs.go#L13-L15

eol/alpine/eof.go
```diff
@@ -0,0 +1,109 @@
...
+	eolDates["edge"] = time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC)
+
+	return eolDates, nil
+}
+
+func (config Config) save(dates map[string]time.Time) error {
+	dir := filepath.Join(config.vulnListDir, eolAlpineFolder)
+	if err := os.RemoveAll(dir); err != nil {
+		return xerrors.Errorf("unable to remove github security advisory directory: %w", err)
+	}
```
knqyf263, line 99: It is a single file, right? Looks like we don't have to remove dir as it overwrites the existing file.

eol/alpine/eof.go
```diff
@@ -0,0 +1,109 @@
...
+func NewConfig() Config {
+	return Config{
+		vulnListDir: utils.VulnListDir(),
+		eolUrl:      eolAlpineUrl,
+		retry:       retry,
+		appFs:       afero.NewOsFs(),
+	}
+}
+
+func (config Config) Update() error {
```
knqyf263, line 40: nit: a single letter name is fine.

```suggestion
func (c Config) Update() error {
```

eol/alpine/eof.go
```diff
@@ -0,0 +1,109 @@
...
+		eolDates[version] = date
+	})
+
+	// edge version doesn't have EOL. Add max date.
+	eolDates["edge"] = time.Date(9999, 1, 1, 0, 0, 0, 0, time.UTC)
+
+	return eolDates, nil
+}
+
+func (config Config) save(dates map[string]time.Time) error {
```
knqyf263, line 95: ditto

eol/alpine/eof.go
```diff
@@ -0,0 +1,109 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+const (
+	eolAlpineFolder = "eol/alpine"
+	eolAlpineFile   = "eol-alpine"
```
knqyf263, line 19: What if we have an entire file name here?
```suggestion
	eolAlpineFile   = "alpine.json"
```

knqyf263 (commented):
In addition, we have to add this updater here.
https://github.com/aquasecurity/vuln-list-update/blob/main/.github/workflows/update.yml

main.go
DmitriyLewen: Thanks, changed

eol/alpine/eof_test.go
DmitriyLewen: Added 2 tests

eol/alpine/eof.go
DmitriyLewen: oh, i didn't see that. Thank you. Removed

eol/alpine/eof.go
DmitriyLewen: Removed

eol/alpine/eof.go
DmitriyLewen: Changed

eol/alpine/eof.go
DmitriyLewen: Changed

eol/alpine/eof.go
DmitriyLewen: Changed

DmitriyLewen:
> In addition, we have to add this updater here.
Added, thanks!

DmitriyLewen:
okay, will work on this and write to you.

DmitriyLewen:
Created PoC - https://gist.github.com/DmitriyLewen/b96589679703425306126d1709e42f41
As you can see - problem with azure/mariner OSes only (https://endoflife.date doesn't have eol dates for them)

knqyf263:
We can manually update EOL for Azure Linux.

eoldates/testdata/happy/eoldates.json
```diff
@@ -0,0 +1,26 @@
...
+  ],
+  "ubuntu": [
+    {
+      "name": "25.04",
+      "eolFrom": "2026-01-17"
+    },
+    {
+      "name": "22.04",
+      "eolFrom": "2027-04-01",
+      "eoesFrom": "2032-04-09"
```
DmitriyLewen, line 23: I was thinking about ubuntu ESM:
where we should handle common and ESM versions.

I decided to use our logic for `vuln-list-update` (keep data as in source). We will handle (split) ESM version in trivy-db.

DmitriyLewen:
@knqyf263 I updated this PR.
Take a look, when you have time.

eoldates/eoldates.go
```diff
@@ -0,0 +1,164 @@
...
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return EOLData{}, xerrors.Errorf("unable to read response: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return EOLData{}, xerrors.Errorf("unexpected status code: %d, body: %s", resp.StatusCode, body)
+	}
```
knqyf263, line 138: Do we want to include the response body in the error?

PR #155, opened 2022-05-18
Integrate support for custom types

julianthome:
The changes introduced in this MR make it possible to specify custom types. This is useful for databases following that are based on the https://gitlab.com/gitlab-org/security-products/gemnasium-db format but support additional packages.

### DIFF
--- a/glad/glad.go
+++ b/glad/glad.go
@@ -34,7 +34,12 @@ type Updater struct {
 	appFs                 afero.Fs
 }
 
-func NewUpdater(alternativeRepoURL string, alternativeRepoBranch string) Updater {
+func NewUpdater(alternativeRepoURL string, alternativeRepoBranch string, customTypes string) Updater {
+	if len(customTypes) > 0 {
+		customTypes := strings.Split(customTypes, ",")
+		supportedTypes = append(supportedTypes, customTypes...)
+	}
+
 	return Updater{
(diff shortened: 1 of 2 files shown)


glad/glad.go
```diff
@@ -34,7 +34,12 @@ type Updater struct {
 	appFs                 afero.Fs
 }
 
-func NewUpdater(alternativeRepoURL string, alternativeRepoBranch string) Updater {
+func NewUpdater(alternativeRepoURL string, alternativeRepoBranch string, customTypes string) Updater {
+	if len(customTypes) > 0 {
+		customTypes := strings.Split(customTypes, ",")
+		supportedTypes = append(supportedTypes, customTypes...)
```
Brcrwilliams, line 40: @julianthome Rather than introduce a new flag, I don't see why we shouldn't add the new types to the `supportedTypes` list? I presume that eventually this data will also be included in advisories-community.

julianthome: Thanks a lot @Brcrwilliams. Yes adding the `alpine` package type makes total sense as a default 👍🏼. I just modified the PR accordingly.

However, having the flag/option to pass custom types may come in handy in the future if we want to integrate advisories about package types that are not officially supported yet. Having a flag would give us some flexibility by enabling us to freely experiment with new package types without having to go through upstream PRs.

PR #161, opened 2022-06-14
fix(rocky): added 2 urls parse

DmitriyLewen:
## Description
Rocky database has been split into 2 urls:
- https://dl.rockylinux.org/vault/rocky
- https://download.rockylinux.org/pub/rocky

Added parsing of both links.

### DIFF
--- a/rocky/rocky.go
+++ b/rocky/rocky.go
@@ -35,6 +35,9 @@ var (
 	defaultArches = []string{"x86_64", "aarch64"}
 
 	releaseRegex = regexp.MustCompile(`\d+.\d+[A-Za-z0-9-.]*`)
+
+	ErrNoUpdateInfoField = xerrors.New("no updateinfo field in the repomd")
+	ErrNoRepomdFile      = xerrors.New("no repomd file")
 )
 
 type options struct {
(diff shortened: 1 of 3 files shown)


rocky/rocky.go
```diff
@@ -137,53 +82,60 @@ func NewConfig(opts ...option) Config {
...
+					}
+					if wasWrite {
+						errataWasWrite = true
+					}
 				}
 			}
 		}
+		// we should know if url doesn't contain vulnerabilities
+		// probably database has been changed
+		if !errataWasWrite {
```
knqyf263, line 110: What is this `errataWasWrite` for? Can't we just return an error?

DmitriyLewen: https://dl.rockylinux.org/vault/rocky has [8.3](http://dl.rockylinux.org/vault/rocky/8.3) and [8.4-RC1](http://dl.rockylinux.org/vault/rocky/8.4-RC1) folders without `updateinfo` field in `repomd.xml`.
https://download.rockylinux.org/pub/rocky has [8.4](http://download.rockylinux.org/pub/rocky/8.4/) and [8.5](http://download.rockylinux.org/pub/rocky/8.5/) folders without `repomd.xml` files.
These folders are returning errors, and I've added skip these folders.
But, for example, if database is moved to another link, we don't see that.
To do this, i added this check: if link (https://dl.rockylinux.org/vault/rocky or https://download.rockylinux.org/pub/rocky) doesn't have advisories - return error.

knqyf263: Got it. Looks like `update` doesn't need to return bool. I refactored a bit. https://github.com/aquasecurity/vuln-list-update/pull/161/commits/9fb03250ccfcaf97f4f9a55ea049ff63affd63d6

knqyf263 (approved):

PR #184, opened 2022-11-23
fix(go-vulndb): skip error, if broken several modules

DmitriyLewen:
## Description
Skip error, if broken several modules. Added list of these modules to log.
If all modules are broken, return an error.

### DIFF
--- a/go-vulndb/govulndb.go
+++ b/go-vulndb/govulndb.go
@@ -83,10 +83,12 @@ func (c VulnDB) Update() error {
 		return xerrors.Errorf("index error: %w", err)
 	}
 
+	var modulesErrors []error
 	for moduleName := range modules {
 		entries, err := c.parseModuleEntries(baseURL, moduleName)
 		if err != nil {
-			return xerrors.Errorf("module entry error: %w", err)
+			modulesErrors = append(modulesErrors, err)
+			continue
 		}
 
 		for _, entry := range entries {
@@ -94,6 +96,17 @@ func (c VulnDB) Update() error {
 			filePath := filepath.Join(c.dir, moduleName, entry.ID+".json")
 			if err = utils.Write(filePath, entry); err != nil {
 				return xerrors.Errorf("file write error: %w", err)
+			}
+		}
+	}
+
+	if len(modulesErrors) > 0 {
+		if len(modules) == len(modulesErrors) { // storage/logic broken
+			return xerrors.Errorf("all modules are broken. One of entry error: %w", modulesErrors[0])
+		} else {
+			log.Println("broken several modules:")
+			for _, err := range modulesErrors {
+				log.Println(err)
 			}
 		}
 	}


go-vulndb/govulndb.go
```diff
@@ -83,10 +83,12 @@ func (c VulnDB) Update() error {
 		return xerrors.Errorf("index error: %w", err)
 	}
 
+	var modulesErrors []error
 	for moduleName := range modules {
 		entries, err := c.parseModuleEntries(baseURL, moduleName)
 		if err != nil {
-			return xerrors.Errorf("module entry error: %w", err)
+			modulesErrors = append(modulesErrors, err)
```
knqyf263, line 90: I think we should handle the more specific error. For example, we may want to accept a 404 error here.
https://github.com/aquasecurity/vuln-list-update/blob/c61f90b2481d6769d46db652fb9ecd667e060db1/go-vulndb/govulndb.go#L137

DmitriyLewen: Done

go-vulndb/govulndb.go
```diff
@@ -83,9 +85,14 @@ func (c VulnDB) Update() error {
 		return xerrors.Errorf("index error: %w", err)
 	}
 
+	var modulesErrors []error
 	for moduleName := range modules {
 		entries, err := c.parseModuleEntries(baseURL, moduleName)
 		if err != nil {
+			if strings.HasSuffix(err.Error(), notFoundError) {
```
knqyf263, line 92: I prefer handling that in `parseModuleEntries`.

go-vulndb/govulndb.go
```diff
@@ -98,6 +105,17 @@ func (c VulnDB) Update() error {
 		}
 	}
 
+	if len(modulesErrors) > 0 {
```
knqyf263, line 108: Now, we handle only a 404 error. We probably don't need the check. We cannot detect the case where all modules are 404, but it is acceptable.

go-vulndb/govulndb.go
DmitriyLewen: Done

go-vulndb/govulndb.go
DmitriyLewen: hm... okay, removed this

knqyf263 (approved):

go-vulndb/govulndb.go
```diff
@@ -2,21 +2,23 @@ package govulndb
 
 import (
 	"encoding/json"
+	"fmt"
+	"golang.org/x/xerrors"
```
afdesk, line 6: It's a linter problem. isn't it?

DmitriyLewen: fixed

go-vulndb/govulndb.go
```diff
@@ -123,6 +125,10 @@ func (c VulnDB) parseModuleEntries(baseURL *url.URL, moduleName string) ([]Entry
 
 	res, err := utils.FetchURL(pkgURL.String(), "", c.retry)
 	if err != nil {
+		if strings.Contains(err.Error(), notFoundError) {
+			log.Println(fmt.Sprintf("module %s not found", moduleName))
```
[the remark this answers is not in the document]
DmitriyLewen: Done(#185)

PR #208, opened 2023-05-26
add alt linux support

ipaqsa:
Add ALT Linux support


knqyf263:
Thanks for your contribution! We add new OS's when the community needs them. Otherwise, it is a big effort to maintain a lot of OSes. We'd see first how much demand there is for ALT Linux. Can we open a discussion?

ipaqsa:
Of course, I am the maintainer of alt linux, the system is popular in our country, and I can support it myself, I have a fork of all your repositories with full support for alt linux, we have our own oval database, and we have packaged the project, I will be very glad if you use me as contributors!

ipaqsa:
Did you mean to open a discussion in the main trivy project ? Which category should I choose ?

> Thanks for your contribution! We add new OS's when the community needs them. Otherwise, it is a big effort to maintain a lot of OSes. We'd see first how much demand there is for ALT Linux. Can we open a discussion?

knqyf263:
>I want to become a contributor and support alt

Does it mean you can keep maintaining ALT Linux implementation in Trivy? It is hard to maintain that on our end, but we'd accept it if you or other maintainers will be maintaining it.

knqyf263:
@ipaqsa Great! It is ok, then. We can accept ALT Linux support.
@afdesk will reach out to you.

knqyf263:
@ipaqsa Did you close the PR?

ipaqsa:
Yes, after conversation with afdesk, we decided that it would be good if I just maintain the fork for now.

knqyf263:
IMO, we can accept it upstream if alt maintainers keep maintaining. Could you explain the reason why the fork is better?

PR #211, opened 2023-05-30
ci: add backup and squash vuln-list repo

afdesk:


knqyf263:
in favor of https://github.com/aquasecurity/vuln-list-update/pull/212

PR #212, opened 2023-06-02
ci: squash `vuln-list` using git-filter-repo

afdesk:
Now `vuln-list` is a large repository that contains about 20'000 commits and 117'000'000 objects.
the total size is about 24Gb.
Github has a limit for repositories - 5Gb.

so we have to squash vuln-list - clone the shallow repository (the last 100 commits), squash and push.

### DIFF
--- a/.github/workflows/squash.yml
+++ b/.github/workflows/squash.yml
@@ -26,7 +26,6 @@ jobs:
     - name: back up vuln-list
       run: |
         cd vuln-list-all
-        git remote set-url origin git@github.com:$GITHUB_OWNER/vuln-list-backup.git
         git push -f https://${GITHUB_TOKEN}@github.com/$GITHUB_OWNER/vuln-list-backup.git main
         cd .. && rm -r -f vuln-list-all
 
--- a/.github/workflows/squash.yml
+++ b/.github/workflows/squash.yml
@@ -12,16 +12,17 @@ jobs:
       GITHUB_OWNER: ${{ github.repository_owner }}
       GITHUB_TOKEN: ${{ secrets.ACCESS_TOKEN }}
     steps:
-    - name: Setup github user email and name
-      run: |
-        git config --global user.email "action@github.com"
-        git config --global user.name "GitHub Action"
-
     - name: clone the whole repo for backup
       uses: actions/checkout@v3
       with:
         repository: ${{ github.repository_owner }}/vuln-list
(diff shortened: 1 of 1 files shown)


.github/workflows/squash.yml
```diff
@@ -0,0 +1,39 @@
...
+        git config --global user.email "action@github.com"
+        git config --global user.name "GitHub Action"
+
+    - name: clone a shallow repo
+      uses: actions/checkout@v3
+      with:
+        repository: ${{ github.repository_owner }}/vuln-list
+        token: ${{ secrets.ACCESS_TOKEN }}
+        path: vuln-list
+        fetch-depth: 100
```
knqyf263, line 31: I think vuln-list push 40-50 commits a day. Can we keep 2,000 commits?

.github/workflows/squash.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    - name: squash and push
+      run: |
+        cd vuln-list
+        git replace -f --graft $(git rev-list --max-parents=0 HEAD)
+        git filter-repo --force
+        git remote add origin https://github.com/afdesk/vuln-list.git
```
knqyf263, line 38: We explicitly specify the remote URL below. Do we need this line?

.github/workflows/squash.yml
```diff
@@ -0,0 +1,39 @@
...
+        git push --force --set-upstream https://${GITHUB_TOKEN}@github.com/afdesk/vuln-list.git main
```
knqyf263, line 39: We have to use `ACCESS_TOKEN`, not `GITHUB_TOKEN`, right?

```suggestion
        git push --force --set-upstream https://${ACCESS_TOKEN}@github.com/aquasecurity/vuln-list.git main
```

knqyf263: Using `github.repository_owner` may be better.

```suggestion
        git push --force --set-upstream https://${GITHUB_TOKEN}@github.com/${{ github.repository_owner }}/vuln-list.git main
```

knqyf263:
And I have a question. Does it replace commit dates? I'd keep the original dates.

knqyf263:
>And I have a question. Does it replace commit dates? I'd keep the original dates.

I confirmed it kept the original dates.

knqyf263 (approved):

PR #222, opened 2023-07-06
chore: remove go-vulndb security advisories

DmitriyLewen:
## Description
GitHub advisory database moved to OSV schema and uses Golang modules for advisories. 
That is why we don't need to use `go-vulndb` now and we can just use `ghsa`.


knqyf263:
Don't we need any changes on GHSA?

DmitriyLewen:
We don't need changes for `ghsa`.
We are already saving advisories for `Go` - https://github.com/aquasecurity/vuln-list/tree/main/ghsa/go

knqyf263 (approved):

PR #223, opened 2023-07-06
refactor(ghsa): use ghsa repository instead of API to get advisories

DmitriyLewen:
## Description
https://github.com/github/advisory-database is open public.
We can get advisories from repository instead of API.

We will also switch to the OSV scheme.

TODO:
- update tests
- add test for advisory with some ecosystems in 1 file (https://github.com/advisories/GHSA-cfgp-2977-2fmm)

### DIFF
--- a/ghsa/ghsa.go
+++ b/ghsa/ghsa.go
@@ -2,187 +2,152 @@ package ghsa
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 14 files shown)


DmitriyLewen:
There is no need to change these.
We simply download https://github.com/github/advisory-database in trivy-db - https://github.com/aquasecurity/trivy-db/pull/345

PR #229, opened 2023-07-25
chore: fall back into vuln-list-reserve

knqyf263:
vuln-list is down now, and it is failing to build trivy-db. We need a temporal solution until it will be back.


knqyf263:
Confirmed it worked.
https://github.com/aquasecurity/vuln-list-update/actions/runs/5653559495/job/15314943724#step:7:100

DmitriyLewen (approved):

PR #232, opened 2023-08-01
feat(osv): add last_affected and limit fields

nikpivkin:
Added last_affected and limit fields. https://ossf.github.io/osv-schema/#affectedrangesevents-fields


knqyf263 (approved):

knqyf263:
I've approved it, but the test is failing.

PR #239, opened 2023-09-19
feat: add k8s cve collector

chen-keinan:
Add k8s upstream CVE updater.
Related https://github.com/aquasecurity/trivy/issues/5020


knqyf263 (commented):
After reviewing MITRE API, it looks almost impossible to cover all edge cases as the affected field is a mess (see examples below), even though our code is complicated.
https://cveawg.mitre.org/api/cve/CVE-2023-2727
https://cveawg.mitre.org/api/cve/CVE-2023-2728
https://cveawg.mitre.org/api/cve/CVE-2022-3294
https://cveawg.mitre.org/api/cve/CVE-2022-3294

Some vulnerabilities don't even exist.
https://cveawg.mitre.org/api/cve/CVE-2022-3172

So, I'd suggest simplifying the logic and open PRs instead of pushing commits. Then, we manually modify and merge them into the main branch. It will not be every day. There are only 43 vulnerabilities so far.

```
$ curl -sL https://k8s.io/docs/reference/issues-security/official-cve-feed/index.json | jq '.items | length'                                                                                                                                                                                                                                                                                                                             43
```

What do you think? @chen-keinan @itaysk

k8s/utils/parser.go
```diff
@@ -0,0 +1,154 @@
...
+		return introduce, to
+	}
+
+	if MinorVersion(introduce) {
+		return introduce + ".0", to
+	}
+
+	if lIndex := strings.LastIndex(to, "."); lIndex != -1 {
+		return strings.TrimSpace(fmt.Sprintf("%s.%s", to[:lIndex], "0")), to
+	}
```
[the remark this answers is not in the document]
chen-keinan: yes because it is type  `lessThen` or `lessThenEqual` , so if `from` is not `0` or `major version only (1.2)` then its `from` vulnerable version start with beginning of version `1.2.0`.

chen-keinan:
> After reviewing MITRE API, it looks almost impossible to cover all edge cases as the affected field is a mess (see examples below), even though our code is complicated. https://cveawg.mitre.org/api/cve/CVE-2023-2727 https://cveawg.mitre.org/api/cve/CVE-2023-2728 https://cveawg.mitre.org/api/cve/CVE-2022-3294 https://cveawg.mitre.org/api/cve/CVE-2022-3294
> 
> Some vulnerabilities don't even exist. https://cveawg.mitre.org/api/cve/CVE-2022-3172
> 
> So, I'd suggest simplifying the logic and open PRs instead of pushing commits. Then, we manually modify and merge them into the main branch. It will not be every day. There are only 43 vulnerabilities so far.
> 
> ```
> $ curl -sL https://k8s.io/docs/reference/issues-security/official-cve-feed/index.json | jq '.items | length'                                                                                                                                                                                                                                                                                                                             43
> ```
> 
> What do you think? @chen-keinan @itaysk

@knqyf263  I think that the code produce very accurate vuln data, I have checked it several times.
I think we should keep it to produce the data and instead of push it automatically it can raise a PR (for manual review)  to make sure that there are no exception.

k8s/utils/parser.go
knqyf263, line 1: This package is dedicated to k8s. Do we need a separate package? Can we move these files under `k8s`?

chen-keinan: sure

chen-keinan:
@knqyf263 I have added examples

knqyf263:
Thanks! I'm taking another look.

k8s/parser.go
```diff
@@ -0,0 +1,150 @@
...
+		return introduce + ".0", to
+	}
+	// Example: https://cveawg.mitre.org/api/cve/CVE-2019-1002100
+	if lIndex := strings.LastIndex(to, "."); lIndex != -1 {
+		return strings.TrimSpace(fmt.Sprintf("%s.%s", to[:lIndex], "0")), to
+	}
+	return introduce, to
+}
+
+func extractRangeVersions(introduce string) (string, string) {
```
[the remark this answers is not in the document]
chen-keinan: I have Added example in code

k8s/k8s.go
```diff
@@ -0,0 +1,207 @@
...
+			events = append(events, osv.Event{Introduced: av.Introduced})
+		}
+		if len(av.Fixed) > 0 {
+			events = append(events, osv.Event{Fixed: av.Fixed})
+		} else if len(av.LastAffected) > 0 && len(av.Fixed) == 0 {
+			events = append(events, osv.Event{LastAffected: av.LastAffected})
+		}
+		if len(av.Introduced) > 0 && len(av.LastAffected) == 0 && len(av.Fixed) == 0 {
+			events = append(events, osv.Event{LastAffected: av.Introduced})
+		}
```
knqyf263, line 135: nit

```suggestion
		if len(av.Fixed) > 0 {
			events = append(events, osv.Event{Fixed: av.Fixed})
		} else if len(av.LastAffected) > 0 {
			events = append(events, osv.Event{LastAffected: av.LastAffected})
		} else if len(av.Introduced) > 0 {
			events = append(events, osv.Event{LastAffected: av.Introduced})
		}
```

k8s/k8s.go
```diff
@@ -0,0 +1,207 @@
...
+		ranges = append(ranges, osv.Range{
+			Events: events,
+		})
+		affected = append(affected, osv.Affected{Ranges: ranges, Package: osv.Package{Name: p, Ecosystem: "k8s"}, Severities: []osv.Severity{{Type: cvss.Vector, Score: fmt.Sprintf("%.1f", cvss.Score)}}})
```
knqyf263, line 139: I'm not sure, but "Kubernetes" might be better.
https://ossf.github.io/osv-schema/#affectedpackage-field

```suggestion
		affected = append(affected, osv.Affected{Ranges: ranges, Package: osv.Package{Name: p, Ecosystem: "Kubernetes"}, Severities: []osv.Severity{{Type: cvss.Vector, Score: fmt.Sprintf("%.1f", cvss.Score)}}})
```

knqyf263 (commented):
We talked offline that we could directly commit initial data, but can we also make PRs for initial data? It would be a good testing for our workflow. If we merge several PRs, trivy-db will not be blocked by review.

PRs should be created per CVE-ID. It allows us to merge some of them.

k8s/k8s.go
```diff
@@ -0,0 +1,206 @@
...
+		k8sComponent = mitreComponent
+	}
+	return strings.ToLower(fmt.Sprintf("%s/%s", upstreamOrgByName(k8sComponent), upstreamRepoByName(k8sComponent)))
+}
+
+func validateCvesData(cves []*osv.OSV) error {
+	var result error
+	for _, cve := range cves {
+		if len(cve.ID) == 0 {
+			result = multierror.Append(result, fmt.Errorf("\nid is mssing on cve #%s", cve.ID))
```
knqyf263, line 157: We no longer need `multierror` because `errors.Join` was introduced in Go 1.20. You can merge [the main branch](https://github.com/aquasecurity/vuln-list-update/commit/e9a5af7af0730c64cfab6aef705ce37c89e1db80), then Go 1.20 will be available.
https://pkg.go.dev/errors#Join

chen-keinan:
@knqyf263 I have add the changes we discussed.
- when the collector will run base on cron it will check if CVE already exist in `vuln-list-k8s` and its modified date is more updated before creating PR for it.
- new shell  script to run from workflow to create pr per CVE but before it will validate that a same pr not exist already

create_pr.sh
```diff
@@ -0,0 +1,50 @@
...
+
+cd "$VULN_LIST_DIR" || exit 1
+
+if [[ -n $(git status --porcelain) ]]; then
+  # List changed files
+CHANGED_FILES=$(git ls-files . --exclude-standard --others | grep "CVE")
+REPO="$REPOSITORY_OWNER/$VULN_LIST_DIR"
+BASE_BRANCH="main"
+# Loop through changed files and create PRs
+for FILE in $CHANGED_FILES; do
```
knqyf263, line 20: Should we indent the lines?

create_pr.sh
```diff
@@ -0,0 +1,50 @@
...
+
+  BRANCH_NAME=$(echo $FILE | tr / -)
+  PR_TITLE="Update $FILE"
+  PR_BODY="This PR updates $FILE"
+
+  # Check if a PR with the same branch name already exists
+  open_pr_count=$(gh pr list --state open --base $BASE_BRANCH --repo $REPO | grep $FILE | wc -l)
```
knqyf263, line 27: nit: just for consistency

```suggestion
  OPEN_PR_COUNT=$(gh pr list --state open --base $BASE_BRANCH --repo $REPO | grep $FILE | wc -l)
```

create_pr.sh
```diff
@@ -0,0 +1,50 @@
...
+
+  if [ "$open_pr_count" -eq 0 ]; then
+    # Create a new branch and push it
+    
+    git checkout -b $BRANCH_NAME
```
knqyf263, line 32: nit: just to suppress linter issues

```suggestion
    git checkout -b "$BRANCH_NAME"
```

create_pr.sh
```diff
@@ -0,0 +1,50 @@
...
+    echo $file
```
knqyf263, line 33:
```suggestion
    echo $FILE
```

k8s/parser.go
```diff
@@ -0,0 +1,150 @@
+package k8s
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/goark/go-cvss/v3/metric"
+
+	version "github.com/aquasecurity/go-pep440-version"
```
knqyf263, line 9: Isn't [this](https://github.com/aquasecurity/go-version/tree/main/pkg/versi) the package we need rather than PEP440?

knqyf263 (commented):
Looks good! i left small comments.

chen-keinan:
@knqyf263 comments has been addressed

knqyf263:
I have one more question. We manually modify changes if needed and merge PRs. So, we should skip opening PRs against the existing advisories, right? Otherwise, our manual changes will be reverted. Is it covered? I saw create_pr.sh skipped open PRs, but didn't find such a case.

k8s/k8s.go
```diff
@@ -0,0 +1,291 @@
...
+			events = append(events, osv.Event{Fixed: av.Fixed})
+		} else if len(av.LastAffected) > 0 {
+			events = append(events, osv.Event{LastAffected: av.LastAffected})
+		} else if len(av.Introduced) > 0 && len(av.LastAffected) == 0 && len(av.Fixed) == 0 {
+			events = append(events, osv.Event{LastAffected: av.Introduced})
+		}
+		ranges = append(ranges, osv.Range{
+			Events: events,
+		})
+		affected = append(affected, osv.Affected{Ranges: ranges, Package: osv.Package{Name: p, Ecosystem: "kubernetes"}, Severities: []osv.Severity{{Type: cvss.Vector, Score: fmt.Sprintf("%.1f", cvss.Score)}}})
```
knqyf263, line 149: The score field must contain the CVSS vector (I know it is confusing).
https://ossf.github.io/osv-schema/#severity-field

```
  "severity": [
    {
      "type": "CVSS_V3",
      "score": "CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:N/A:N"
    }
```

knqyf263 (commented):
I found one more mistake.

k8s/k8s.go
chen-keinan: fixed.

chen-keinan:
Fixed

knqyf263:
I'm sorry if I'm missing something, but is it covered somewhere?
https://github.com/aquasecurity/vuln-list-update/pull/239#issuecomment-1735320862

chen-keinan:
> I'm sorry if I'm missing something, but is it covered somewhere? [#239 (comment)](https://github.com/aquasecurity/vuln-list-update/pull/239#issuecomment-1735320862)

yes its covered in code , follow this [function](https://github.com/aquasecurity/vuln-list-update/blob/166257a854438736fe0240ecb0ac809bfa108419/k8s/k8s.go#L272)

k8s/k8s.go
```diff
@@ -0,0 +1,291 @@
...
+
+func cveMissingImportantData(vulnerability *Cve) bool {
+	return len(vulnerability.versions) == 0 ||
+		len(vulnerability.Package) == 0 ||
+		len(vulnerability.CvssV3.Vector) == 0
+}
+
+// getExitingCvesToModifiedMap get the existing cves from vuln-list-k8s repo and map it to cve id and last updated
+func getExitingCvesToModifiedMap() (map[string]string, error) {
+	response, err := http.Get(vulnListRepoTarBall)
```
knqyf263, line 221: It is already [stored](https://github.com/aquasecurity/vuln-list-update/blob/166257a854438736fe0240ecb0ac809bfa108419/.github/workflows/k8s.yml#L24-L29) locally. Do we need to download it again?

k8s/k8s.go
```diff
@@ -0,0 +1,291 @@
...
+		if err != nil {
+			return false
+		}
+		// check if the current collcted cve is older or same as the existing one
+		if currentLastUpdated.Before(existastUpdated) || currentLastUpdated == existastUpdated {
+			return true
+		}
+	}
+	return false
+}
```
knqyf263, line 291: nit: I prefer aligning the happy path to the left edge.
https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88


```suggestion
	if lastUpdated, ok := existCveLastUpdated[cveID]; !ok {
		return false
	}
	
	existastUpdated, err := time.Parse(time.RFC3339, lastUpdated)
	if err != nil {
		return false
	}
	currentLastUpdated, err := time.Parse(time.RFC3339, currentCVEUpdated)
	if err != nil {
		return false
	}
	// check if the current collcted cve is older or same as the existing one
	if currentLastUpdated.Before(existastUpdated) || currentLastUpdated == existastUpdated {
		return true
	}
	return false
}
```

k8s/k8s.go
```diff
@@ -0,0 +1,291 @@
...
+	return cveTimeMap, nil
+}
+
+// olderCve check if the current cve is older than the existing one
+func olderCve(cveID string, currentCVEUpdated string, existCveLastUpdated map[string]string) bool {
+	if len(existCveLastUpdated) == 0 {
+		return false
+	}
+	if lastUpdated, ok := existCveLastUpdated[cveID]; ok {
+		existastUpdated, err := time.Parse(time.RFC3339, lastUpdated)
```
knqyf263, line 277: typo?
```suggestion
		existLastUpdated, err := time.Parse(time.RFC3339, lastUpdated)
```

k8s/k8s.go
chen-keinan: I would prefer to not couple between code logic to github action, wdyt ?

knqyf263: Agree. I had the same feeling, but it already depends on the git dir. If vuln-list-k8s doesn't exist, it will fail. I'm suggesting the reuse of that dir.
https://github.com/aquasecurity/vuln-list-update/blob/2927f2b673cd17083870806ed120cffd80cada41/k8s/k8s.go#L83

`vuln-list-update` assumes the vuln-list dir exists. The dir path can be passed from outside.
https://github.com/aquasecurity/vuln-list-update/blob/2927f2b673cd17083870806ed120cffd80cada41/main.go#L43

What if writing Makefile like trivy-db? The benefit is the source code can be much simpler.
https://github.com/aquasecurity/trivy-db/blob/b7e0a5886daa539ee175c666ecf051c2da94ef05/Makefile#L91-L100

chen-keinan: I get your point, you convinced me , I think it will be simpler , I'll implement it with existing as suggested initially

chen-keinan: fixed.

knqyf263 (approved):
LGTM!

knqyf263:
Triggered. Let's see how it goes.
https://github.com/aquasecurity/vuln-list-update/actions/runs/6326525504

knqyf263:
@chen-keinan Failed.

>2023/09/27 13:19:48 Chainguard update error: error in k8s update: invalid character '#' looking for beginning of value

k8s/k8s.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	cvesMap, err := getExitingCvesToModifiedMap()
+	if err != nil {
+		return nil, err
+	}
+	return ParseVulnDBData(db, cvesMap)
+}
+
+const (
+	// excludeNonCoreComponentsCves exclude  cves with missing data or non k8s core components
```
itaysk, line 63: can you please elaborate on this?

chen-keinan: this cves which are on non core components, we do not find them with kbom

itaysk: if we don't find them, why do we care if they're in the db?

chen-keinan: its just matter of keeping the db relevant to our needs and purpose. not more then that

PR #245, opened 2023-09-28
build: replace PAT with ORG_REPO_TOKEN 

knqyf263:
## Description
ACCESS_TOKEN is my personal access token. It should be replaced with an organization secret.
Also, the script currently doesn't exit even if a command fails. This PR add `-eu` so the failure can stop the script.


chen-keinan (approved):

knqyf263:
I'd defer to @DmitriyLewen. He might find any problems as he's a more careful man than I am 😄

DmitriyLewen (approved):
LGTM.


1 point: you moved scripts to separate folder.
I guess we can remove `push.sh`? We don't seem to use them.
But maybe we need to do it in another PR.

knqyf263:
Thanks, @DmitriyLewen. I've removed the script.

PR #248, opened 2023-10-01
chore: increase k8s open prs limit

chen-keinan:
increase `gh list pr` page limit when query for open PRs
fix this [error](https://github.com/aquasecurity/vuln-list-update/actions/runs/6368520330/job/17287946530) 

I have increased `limit to 50` (enough for now , default limit is 30) should be ok for now I assume it will happen only for initial db creation

### DIFF
--- a/scripts/create_pr.sh
+++ b/scripts/create_pr.sh
@@ -24,7 +24,7 @@ if [[ -n $(git status --porcelain) ]]; then
     PR_BODY="This PR updates $FILE"
 
     # Check if a PR with the same branch name already exists
-    OPEN_PR_COUNT=$(gh pr list --state open --base $BASE_BRANCH --repo "$REPO" | grep "$FILE" | wc -l)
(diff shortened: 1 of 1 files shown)


chen-keinan:
@knqyf263 add this small pr to script not to fail, please have a look

knqyf263:
It looks like another error, doesn't it?

>a pull request for branch "upstream-CVE-2017-1002101.json" into branch "main" already exists:

chen-keinan:
> It looks like another error, doesn't it?
> 
> > a pull request for branch "upstream-CVE-2017-1002101.json" into branch "main" already exists:

the problem is when it run :
`gh pr list --state open --base main --repo github.com/aquasecurity/vuln-list-k8s | grep upstream-CVE-2017-1002101.json | wc -l`

it return `0` ,  because the default limit is `30` and the repo has 37 prs and the PR `upstream-CVE-2017-1002101.json` come only with 2nd page. the script should skip and do not try to open a new PR

if we increase the limit to `50` the query above will return `1` and it will skip and do not try to create another PR

knqyf263:
Thanks for explaining. I got it.

knqyf263 (approved):

PR #249, opened 2023-10-02
fix: aggregate k8s events under same package

chen-keinan:
fix: aggregate k8s events under same package


k8s/k8s.go
```diff
@@ -191,7 +201,10 @@ func validateCvesData(cves []*osv.OSV) error {
 					}
 				}
 				for _, r := range v.Ranges {
-					for i := 1; i < len(r.Events); i++ {
+					if len(r.Events)%2 != 0 {
```
knqyf263, line 204: The events aren't necessarily a pair in the OSV spec.

```
"ranges": [ {
    "type": "SEMVER",
    "events": [
      { "introduced": "0" },
    ]
} ]
```

cf. https://ossf.github.io/osv-schema/#unfixed-vulnerability-example

But it's okay to have this validation if it is a requirement of our script. @chen-keinan Do you think our k8s advisories must be a pair?

chen-keinan: I can remove this validation as anyway we are doing PR review manually

chen-keinan:
@knqyf263 please have another look

knqyf263 (commented):
Looks good to me. Can you please run `check-jsonschema` against the generated advisories? @chen-keinan 
https://github.com/aquasecurity/vuln-list-k8s/pull/95

chen-keinan:
> Looks good to me. Can you please run `check-jsonschema` against the generated advisories? @chen-keinan ref. [aquasecurity/vuln-list-k8s#95](https://github.com/aquasecurity/vuln-list-k8s/pull/95)

I have added it to create_pr script and not as a separate action

knqyf263:
>I have added it to create_pr script and not as a separate action

Ohhhhh, I'm sorry I didn't have enough words. I just asked you to run `check-jsonschema` on your computer to make sure the advisories are valid now.

chen-keinan:
@knqyf263 you think its not a must to have a schema validation here ? I assume it can only do good :)

knqyf263:
You can decide it. The JSON will be anyway validated on PRs in vuln-list-k8s. 
I have a small concern we need to update two places when the spec is updated to v1.7.0, but it is no big deal.

PR #254, opened 2023-11-09
feat(debian/tracker): add TEMP entry instead of CVE-yyyy-XXXX

MaineK00n:
There are multiple entries called `CVE-yyyy-XXXX` in `CVE/list`.
ref. https://salsa.debian.org/search?search=XXXX&nav_source=navbar&project_id=555&group_id=2102&search_code=true&repository_ref=master

However, the current vuln-list-update does not consider them, and only one is saved as `vuln-list-debian/tracker/CVE/yyyy/CVE-yyyy-XXXX.json`.
ref. https://github.com/search?q=repo%3Aaquasecurity%2Fvuln-list-debian%20XXXX&type=code

If you look at these entries on the Debian Security Tracker on the web, they will be converted to entries called TEMP.
ref. https://security-tracker.debian.org/tracker/data/fake-names

The conversion method is as follows.
https://salsa.debian.org/security-tracker-team/security-tracker/-/blob/50ca55fb66ec7592f9bc1053a11dbf0bd50ee425/lib/python/bugs.py#L402-408

This PR changes to save to `vuln-list-debian/tracker/CVE/TEMP/TEMP-{Debian Bug Number}-{md5sum(description)}.json` in case of `CVE-yyyy-XXXX`.

### DIFF
--- a/debian/tracker/debian.go
+++ b/debian/tracker/debian.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/md5"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -158,8 +159,26 @@ func (c Client) update(dirname string, bugs []Bug) error {
 	for _, bug := range bugs {
 		dir := filepath.Join(c.vulnListDir, trackerDir, dirname)
 		if dirname == "CVE" {
-			if err := utils.SaveCVEPerYear(dir, bug.Header.ID, bug); err != nil {
-				return xerrors.Errorf("debian: failed to save CVE per year: %w", err)
+			if strings.HasSuffix(bug.Header.ID, "-XXXX") {
+				var bugno int
+				for _, ann := range bug.Annotations {
+					if ann.Type == "package" && ann.BugNo != 0 {
+						bugno = ann.BugNo
+						break
+					}
+				}
+
+				bug.Header.ID = tempBugName(bugno, bug.Header.Description)
+
+				fileName := fmt.Sprintf("%s.json", bug.Header.ID)
+				filePath := filepath.Join(dir, "TEMP", fileName)
+				if err := utils.Write(filePath, bug); err != nil {
+					return xerrors.Errorf("debian: write error (%s): %w", filePath, err)
+				}
+			} else {
+				if err := utils.SaveCVEPerYear(dir, bug.Header.ID, bug); err != nil {
+					return xerrors.Errorf("debian: failed to save CVE per year: %w", err)
+				}
 			}
 		} else {
 			fileName := fmt.Sprintf("%s.json", bug.Header.ID)
@@ -350,4 +369,16 @@ func (c Client) parseSources(sourcePath string) ([]textproto.MIMEHeader, error) 
 	}
 
 	return headers, nil
+}
+
+// ref. https://salsa.debian.org/security-tracker-team/security-tracker/-/blob/50ca55fb66ec7592f9bc1053a11dbf0bd50ee425/lib/python/bugs.py#L402
+func tempBugName(bugNumber int, description string) string {
+	switch {
+	case strings.HasPrefix(description, "["):
+		description = strings.TrimPrefix(strings.TrimSuffix(description, "]"), "[")
+	case strings.HasPrefix(description, "("):
+		description = strings.TrimPrefix(strings.TrimSuffix(description, ")"), "(")
+	}
+	hash := fmt.Sprintf("%x", md5.Sum([]byte(description)))
+	return fmt.Sprintf("TEMP-%07d-%s", bugNumber, strings.ToUpper(hash[:6]))
 }
--- a/debian/tracker/debian_test.go
+++ b/debian/tracker/debian_test.go
@@ -154,6 +154,64 @@ func TestClient_Update(t *testing.T) {
 						},
 					},
 				},
+				filepath.Join("CVE", "TEMP", "TEMP-1053115-9454E3.json"): {
+					Header: &tracker.Header{
+						Original:    "CVE-2023-XXXX [code execution via malformed XTGETTCAP]",
+						ID:          "TEMP-1053115-9454E3",
+						Description: "[code execution via malformed XTGETTCAP]",
+					},
+					Annotations: []*tracker.Annotation{
+						{
+							Original: "- foot 1.15.3-2 (bug #1053115)",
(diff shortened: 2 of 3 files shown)


debian/tracker/debian.go
```diff
@@ -351,3 +370,15 @@ func (c Client) parseSources(sourcePath string) ([]textproto.MIMEHeader, error)
...
 }
+
+// ref. https://salsa.debian.org/security-tracker-team/security-tracker/-/blob/50ca55fb66ec7592f9bc1053a11dbf0bd50ee425/lib/python/bugs.py#L402
+func tempBugName(bugNumber int, description string) string {
+	switch {
+	case strings.HasPrefix(description, "["):
+		description = strings.TrimPrefix(strings.TrimSuffix(description, "]"), "[")
+	case strings.HasPrefix(description, "("):
+		description = strings.TrimPrefix(strings.TrimSuffix(description, ")"), "(")
+	}
```
DmitriyLewen, line 381: Are there cases when description only contains suffix `]` (`)`) and doesn't contain prefix `[`(`(`)?
if not - it looks like we can remove `switch`.

k8s/testdata/expected-vulndb.json
```diff
@@ -1 +1 @@
...ixed":"1.9.4"}]}]}],"references":[{"type":"ADVISORY","url":"https://github.com/kubernetes/kubernetes/issues/60814"},{"type":"ADVISORY","url":"https://www.cve.org/cverecord?id=CVE-2017-1002102"}]},{"id":"CVE-2017-1002101","modified":"2018-03-05T20:53:58Z","published":"2018-03-05T20:53:58Z","summary":"subpath volume mount handling allows arbitrary file access in host filesystem","details":"In Kubernetes versions 1.3.x, 1.4.x, 1.5.x, 1.6.x and prior to versions 1.7.14, 1.8.9 and 1.9.4 containers using subpath volume mounts with any volume type (including non-privileged pods, subject to file permissions) can access files/directories outside of the volume, including the host's filesystem.","affected":[{"package":{"ecosystem":"kubernetes","name":"k8s.io/kubernetes"},"severity":[{"type":"CVSS_V3","score":"CVSS:3.0/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:H"}],"ranges":[{"type":"SEMVER","events":[{"introduced":"1.3.0"},{"fixed":"1.7.14"},{"introduced":"1.8.0"},{"fixed":"1.8.9"},{"introduced":"1.9.0"},{"fixed":"1.9.4"}]}]}],"references":[{"type":"ADVISORY","url":"https://github.com/kubernetes/kubernetes/issues/60813"},{"type":"ADVISORY","url":"https://www.cve.org/cverecord?id=CVE-2017-1002101"}]}]
```
DmitriyLewen, line 1: Are these expected changes?

DmitriyLewen (commented):
@MaineK00n Thanks for great work!
I left 2 small comment. 
Take a look, when you have time, please.

debian/tracker/debian.go
```diff
@@ -158,8 +159,26 @@ func (c Client) update(dirname string, bugs []Bug) error {
...
+					}
+				}
+
+				bug.Header.ID = tempBugName(bugno, bug.Header.Description)
+
+				fileName := fmt.Sprintf("%s.json", bug.Header.ID)
+				filePath := filepath.Join(dir, "TEMP", fileName)
+				if err := utils.Write(filePath, bug); err != nil {
+					return xerrors.Errorf("debian: write error (%s): %w", filePath, err)
+				}
```
DmitriyLewen, line 177: Maybe we want to move this to separate function?

debian/tracker/debian.go
MaineK00n: First, when calculating the hash of description with md5, it seems necessary to remove the parentheses at both ends.

Referring to the implementation below, if the first character of the description has either `(` or `[`, an error will occur if the last character is not closed with the corresponding parentheses such as `)` or `]`. 
https://salsa.debian.org/security-tracker-team/security-tracker/-/blob/778d44678936649eb550e589645c60a50da81f88/lib/python/sectracker/parsers.py#L308-313

In other words, the case you mentioned does not seem to exist.

k8s/testdata/expected-vulndb.json
MaineK00n: No, the whole CI test fails, so I just fixed it and it has nothing to do with the changes to the debian tracker.

debian/tracker/debian.go
MaineK00n: The logic as a whole is not that complicated, and the part that generates a new ID using the bug number and description has already been extracted into a function, so I don't think any more is needed.

debian/tracker/debian.go
MaineK00n: For example, if the description is `[description()]`, if you randomly delete `[(` from the prefix and `)]` from the suffix, it will become `description(`).
In this case, we expect it to be `description()`.
For this purpose, I adopted an implementation that looks at the type of parenthesis in the first character and determines the type of parenthesis to be removed from the prefix and suffix.

k8s/testdata/expected-vulndb.json
MaineK00n: https://github.com/aquasecurity/vuln-list-update/actions/runs/6817001043/job/18539561269

debian/tracker/debian.go
DmitriyLewen: I suspected there might be something like this.
Thanks for example.

k8s/testdata/expected-vulndb.json
DmitriyLewen: Give me some time to understand why k8s test broke.

DmitriyLewen: Thanks for information.
I think we don't need to merge this change with your changes.
It's also strange that test stops working correctly without changes in `k8s` package.

I create #255 for this case.

MaineK00n: I dropped the commit that made the changes to this k8s.

DmitriyLewen: https://github.com/aquasecurity/vuln-list-update/pull/255 has been merged

MaineK00n: I rebased the main branch.

DmitriyLewen (approved):

PR #256, opened 2023-11-13
test(k8s): remove internet access

DmitriyLewen:
## Description
`Test_ParseVulneDB` calls Internet access.
The result of this test is subject to change.

We need to use test http server for this test.


DmitriyLewen:
@chen-keinan Hello!

I updated `Test_ParseVulneDB` for `k8s` package.
I removed some CVEs from `k8s-db.json`, but i tried to add CVE for each case from `ParseVulnDBData` and `parseMitreCve` ( i don't check cases from `sanitizedVersions`, because there is `TestSanitizedVersion`).

Take a look these changes, please, when you have time.

k8s/k8s_test.go
```diff
@@ -14,7 +19,26 @@ func Test_ParseVulneDB(t *testing.T) {
...
+		if !d.IsDir() {
+			f, err := os.ReadFile(path)
+			assert.NoError(t, err)
+			mux.HandleFunc(fmt.Sprintf("/%s", filepath.Base(path)), func(w http.ResponseWriter, r *http.Request) {
+				_, err = w.Write(f)
+				assert.NoError(t, err)
+			})
+		}
+		return nil
+	})
```
chen-keinan, line 34: nit: `err error` and `r *http.Request` are not in use 
```suggestion
	err = filepath.WalkDir("./testdata/mitreCVEs", func(path string, d fs.DirEntry, err error) error {
		if !d.IsDir() {
			var f []byte
			f, err = os.ReadFile(path)
			assert.NoError(t, err)
			mux.HandleFunc(fmt.Sprintf("/%s", filepath.Base(path)), func(w http.ResponseWriter, _ *http.Request) {
				_, err = w.Write(f)
				assert.NoError(t, err)
			})
		}
		return nil
	})
```

chen-keinan (commented):
lgtm I have added on `nit` comment

k8s/k8s_test.go
```diff
@@ -14,7 +19,27 @@ func Test_ParseVulneDB(t *testing.T) {
...
+	mux := http.NewServeMux()
+	err = filepath.WalkDir("./testdata/mitreCVEs", func(path string, d fs.DirEntry, err error) error {
+		if !d.IsDir() {
+			var f []byte
+			f, err = os.ReadFile(path)
+			assert.NoError(t, err)
+			mux.HandleFunc(fmt.Sprintf("/%s", filepath.Base(path)), func(w http.ResponseWriter, _ *http.Request) {
+				_, err = w.Write(f)
+				assert.NoError(t, err)
+			})
```
knqyf263, line 32: We might be able to leverage `http.FileServer`.
https://pkg.go.dev/net/http#FileServer

If you need to modify the URL's path, `http.StripPrefix` can help.
https://pkg.go.dev/net/http#StripPrefix

knqyf263 (approved):

k8s/k8s_test.go
DmitriyLewen: Thanks for the advice!

PR #258, opened 2023-11-27
refactor(nvd): use API instead of JSON feeds

DmitriyLewen:
## Description
Use `NVD API` instead of `NVD Data Feeds`.

Test run - https://github.com/DmitriyLewen/vuln-list-update/actions/runs/7097735967/job/19318355668
Test repo - https://github.com/dmitriylewen/vuln-list-nvd

## Related Issues:
- aquasecurity/trivy/issues/5658


nvd/nvd.go
```diff
@@ -1,162 +1,250 @@
...
+			continue
+		case http.StatusServiceUnavailable, http.StatusRequestTimeout, http.StatusBadGateway, http.StatusGatewayTimeout:
+			log.Printf("NVD API is unstable: %s. Try to fetch URL again", resp.Status)
+			// NVD API works unstable
+			time.Sleep(1 * time.Second)
+			continue
+		case http.StatusOK:
+			return resp.Body, nil
+		default:
+			return nil, xerrors.Errorf("unexpected status code: %s", resp.Status)
```
DmitriyLewen, line 196: The NVD API is unstable.
I added handling of all possible errors

nvd/nvd.go
```diff
@@ -1,162 +1,250 @@
...
 
-func fetchLastModifiedDate(feed string) (time.Time, error) {
-	log.Printf("Fetching NVD metadata(%s)...\n", feed)
-
-	url := fmt.Sprintf("%s/nvdcve-1.1-%s.meta", baseURL, feed)
-	res, err := utils.FetchURL(url, "", 5)
+func (u Updater) getEntry(url string) (Entry, error) {
+	var entry Entry
+	r, err := fetchURL(url, u.apiKey, u.retry)
 	if err != nil {
```
nikpivkin, line 151: We need to defer closing the body after checking the error

nikpivkin, line 151: We need to defer closing the body after checking the error

nvd/nvd.go
```diff
@@ -1,162 +1,250 @@
...
-		line := scanner.Text()
-		s := strings.SplitN(line, ":", 2)
-		if len(s) != 2 {
+	return entry, xerrors.Errorf("unable to get entry from %q", url)
+}
+
+func fetchURL(url, apiKey string, retry int) (io.ReadCloser, error) {
+	var c http.Client
+	for i := 0; i <= retry; i++ {
+		req, err := http.NewRequest("GET", url, nil)
```
nikpivkin, line 167:
```suggestion
		req, err := http.NewRequest(http.MethodGet, url, nil)
```

nvd/nvd.go
```diff
@@ -1,162 +1,250 @@
...
 		}
-		if s[0] == "lastModifiedDate" {
-			t, err := time.Parse(time.RFC3339, s[1])
-			if err != nil {
-				return time.Time{}, err
-			}
-			return t, nil
+		switch resp.StatusCode {
+		case http.StatusForbidden:
+			log.Printf("NVD rate limit. Waiting to gain access")
```
nikpivkin, line 182:
```suggestion
			log.Println("NVD rate limit. Waiting to gain access")
```

nvd/nvd.go
```diff
@@ -1,162 +1,250 @@
...
+			// NVD limits:
+			// Without API key: 5 requests / 30 seconds window
+			// With API key: 50 requests / 30 seconds window
+			time.Sleep(30 * time.Second)
+			continue
+		case http.StatusServiceUnavailable, http.StatusRequestTimeout, http.StatusBadGateway, http.StatusGatewayTimeout:
+			log.Printf("NVD API is unstable: %s. Try to fetch URL again", resp.Status)
```
nikpivkin, line 189:
```suggestion
			log.Printf("NVD API is unstable: %s. Try to fetch URL again\n", resp.Status)
```

nvd/nvd.go
```diff
@@ -1,162 +1,250 @@
...
+
+func fetchURL(url, apiKey string, retry int) (io.ReadCloser, error) {
+	var c http.Client
+	for i := 0; i <= retry; i++ {
+		req, err := http.NewRequest("GET", url, nil)
+		if err != nil {
+			return nil, xerrors.Errorf("unable to build request for %q: %w", url, err)
+		}
+		if apiKey != "" {
+			req.Header.Set("apiKey", apiKey)
```
nikpivkin, line 172: API key must be included as a URL parameter 
https://nvd.nist.gov/general/news/API-Key-Announcement
>Immediately after activation, API keys can be included as a parameter of the requestor’s URL string.

nvd/nvd.go
```diff
@@ -1,162 +1,250 @@
...
-				return time.Time{}, err
-			}
-			return t, nil
+		switch resp.StatusCode {
+		case http.StatusForbidden:
+			log.Printf("NVD rate limit. Waiting to gain access")
+			// NVD limits:
+			// Without API key: 5 requests / 30 seconds window
+			// With API key: 50 requests / 30 seconds window
+			time.Sleep(30 * time.Second)
```
nikpivkin, line 186: By the way, there is a package `https://pkg.go.dev/golang.org/x/time/rate#NewLimiter` that provides Limiter.

Example of use:
```go
rate.NewLimiter(rate.Every(30*time.Second), 50) // with key
```

nvd/nvd.go
DmitriyLewen: Thanks! Changed in https://github.com/aquasecurity/vuln-list-update/pull/258/commits/763aac6a94019dd88489f0631e2ca63ee4ccd7fb

nvd/nvd.go
DmitriyLewen: Changed in https://github.com/aquasecurity/vuln-list-update/pull/258/commits/763aac6a94019dd88489f0631e2ca63ee4ccd7fb

nvd/nvd.go
DmitriyLewen: log.Printf contains line break - https://go.dev/play/p/ZUxn1hligAt?v=goprev

nvd/types.go
```diff
@@ -0,0 +1,175 @@
+package nvd
+
+// Entry is based on https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema
+type Entry struct {
```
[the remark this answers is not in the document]
DmitriyLewen: I tried using `quicktype` , but result structures were for `definitions` and `properties`. Also `quicktype` created `Format` struct for strings and didn't find any schemes used  (like `cvss-v2.0.json`).
I decided that this was an inconvenient utility for our case.

So i created `Entry` from real response.
But after work with update for Trivy-db i realized that it is not full struct.

After that i updated this struct with using https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema, https://csrc.nist.gov/schema/nvd/api/2.0/external/cvss-v3.1.json, https://csrc.nist.gov/schema/nvd/api/2.0/external/cvss-v3.0.json and https://csrc.nist.gov/schema/nvd/api/2.0/external/cvss-v2.0.json.
 manually.

I tried to keep the order from `cve_api_json_2.0.schema` to easier review (but i moved some fields for `Cve` to reproduce order of response (e.g. referents to end of struct).

nvd/nvd.go
```diff
@@ -99,39 +104,19 @@ func (u Updater) saveEntry(interval timeInterval, startIndex int) (int, error) {
...
 		}
 	}
-	// we need to save LastModEndDate to avoid saving wrong time if we get errors when saving all CVEs (first run)
-	// after this we can switch to entry.Timestamp
-	timestamp, err := time.Parse(nvdTimeFormat, lastModEndDate)
-	if err != nil {
-		return xerrors.Errorf("unable to parse timestamp: %w", err)
-	}
-	// update the Last_updated.json file after saving each entry
-	// to avoid overwriting this entry if we fail to save the next entry
```
knqyf263: Since we set `-eu`, the files are not committed when API calls fail in the middle.
https://github.com/aquasecurity/vuln-list-update/blob/8b9bdf21bcb9235c003a32dd91bc1f04e772803f/scripts/update.sh#L1

Therefore, we assume all requests are done successfully and update `last_updated.json` once. Please let me know if I'm missing something.

nvd/nvd.go
```diff
@@ -88,18 +88,9 @@ func (u Updater) Update() error {
...
-		if err != nil {
-			return xerrors.Errorf("unable to save entry CVEs for %q: %w", interval, err)
-		}
-
-		// if there are more records than maxResultsPerPage
-		// check next records page by page (first page is saved above)
-		for startIndex := u.maxResultsPerPage; startIndex < totalResults; startIndex += u.maxResultsPerPage {
-			_, err = u.saveEntry(interval, startIndex)
-			if err != nil {
+		totalResults := 1 // Set a dummy value to start the loop
```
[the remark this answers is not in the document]
DmitriyLewen: This is good solution! Thanks!

nvd/nvd.go
DmitriyLewen: Oh... I only tested updating `last_updated.json` file on local PC without committing.

I added this for the following case:
This is the first run, we parsed first half (as an example) of all advisories and we get error.
To avoid parsing parsed advisories, I added the update after each interval.
But we try to check all possible cases, also NVD is working on stability of its API.

This way we can only update `last_updated.json` if run is successful.

knqyf263: Yes, we can assume NVD API is stable on its first run. If it doesn't work, let's go with your approach.

knqyf263 (approved):

PR #265, opened 2024-01-19
chore: remove ghsa

DmitriyLewen:
## Description
We use GitHub repository to get list of advisories in Trivy-db - https://github.com/aquasecurity/trivy-db/pull/345
So we don't need to store ghsa advisories in `vuln-list`.


knqyf263:
@tonaim Can we migrate GHSA from API to the repository?

knqyf263 (approved):

PR #268, opened 2024-02-12
include advisories for amazon linux extras

### LINKED ISSUE #267: advisories for amazonlinux extras are missing
In AmazonLinux2 there are a few packages, including some kernel versions, which can be installed via the ["extras" mechanism](https://aws.amazon.com/amazon-linux-2/faqs/#Amazon_Linux_Extras), and that happens through individual mirrors which are separate from the one which is currently used to fetch the update advisories.
This means that some advisories like [this one](https://alas.aws.amazon.com/AL2/ALASKERNEL-5.10-2024-049.html) as currently not present in the vuln-list output.

The list of packages included in the extras can be pulled from
```
https://cdn.amazonlinux.com/2/extras-catalog-${arch}.json
```
with each mirror link then being found at
```
https://cdn.amazonlinux.com/2/extras/${extras_name}/latest/${arch}/mirror.list
```

As far as I am aware, this only applies to Amazon Linux 2 at the time of writing.

piax93:
This fixes #267.

The implementation is not terribly fast, as it iterates through a bunch of URLs. On a very standard home internet connection it still completes in just about under a minute, which seems reasonable for something which is going to run just a few times a day. I can work on parallelizing it if you are aiming for a bit higher performance target.


amazon/amazon.go
```diff
@@ -80,27 +87,56 @@ func (ac Config) update(version, url string) error {
...
+	baseURL := fmt.Sprintf("%s://%s", parsedURL.Scheme, parsedURL.Host)
+	for _, topic := range extrasListing.Topics {
+		url := fmt.Sprintf("%s/%s/extras/%s/latest/x86_64/mirror.list", baseURL, version, topic.Name)
+		mirrorURLs = append(mirrorURLs, url)
+	}
+	return mirrorURLs, nil
 }
 
-func fetchUpdateInfoAmazonLinux(mirrorListURL string) (uinfo *UpdateInfo, err error) {
+func fetchUpdateInfoAmazonLinux(mirrorListURL string, isOptional bool) (uinfo *UpdateInfo, err error) {
```
DmitriyLewen, line 139:
```suggestion
func fetchUpdateInfoAmazonLinux(mirrorListURL string, isExtras bool) (uinfo *UpdateInfo, err error) {
```

amazon/amazon.go
```diff
@@ -122,7 +158,9 @@ func fetchUpdateInfoAmazonLinux(mirrorListURL string) (uinfo *UpdateInfo, err er
 
 		updateInfoPath, err := fetchUpdateInfoURL(u.String())
 		if err != nil {
-			log.Printf("Failed to fetch updateInfo URL: %s\n", err)
+			if !isOptional {
```
DmitriyLewen, line 161: Let's write why we can skip this error.
```suggestion
                        // There are cases when extras don't contain an `updateinfo` value.
			if !isExtras {
```

amazon/amazon.go
```diff
@@ -134,6 +172,9 @@ func fetchUpdateInfoAmazonLinux(mirrorListURL string) (uinfo *UpdateInfo, err er
 		}
 		return uinfo, nil
 	}
+	if isOptional {
```
DmitriyLewen, line 175: same

amazon/amazon.go
```diff
@@ -32,26 +33,32 @@ var (
 		"2022": "https://cdn.amazonlinux.com/al2022/core/mirrors/latest/x86_64/mirror.list",
 		"2023": "https://cdn.amazonlinux.com/al2023/core/mirrors/latest/x86_64/mirror.list",
 	}
+	extrasListURI = map[string]string{
+		"2": "https://cdn.amazonlinux.com/2/extras-catalog-x86_64.json",
```
DmitriyLewen, line 37: Can you add a comment where you found this link?
It may help to add links for new releases next time.

amazon/amazon.go
```diff
@@ -80,27 +87,56 @@ func (ac Config) update(version, url string) error {
...
+		return xerrors.Errorf("failed to fetch extras mirrors: %w", err)
 	}
 
-	bar := pb.StartNew(len(vulns.ALASList))
-	for _, alas := range vulns.ALASList {
-		filePath := filepath.Join(dir, fmt.Sprintf("%s.json", alas.ID))
-		if err = utils.Write(filePath, alas); err != nil {
-			return xerrors.Errorf("failed to write Amazon CVE details: %w", err)
+	bar := pb.StartNew(0)
+	for i, url := range append([]string{url}, extrasMirrors...) {
```
DmitriyLewen, line 96: You overwrite `url` variable.
Can you fix this?

DmitriyLewen (commented):
Hello @piax93 
Thanks for your work.

Look good.
I left small comments. Take a look, please.

DmitriyLewen (approved):
Thanks for your work @piax93 
LGTM.

@knqyf263 I approved this PR. Take a look, when you have time, please.

knqyf263:
How do we distinguish between regular packages and extras packages? I think we need to apply the extras advisories only to extras packages.

piax93:
@knqyf263 mmm, that's a good point, as many package names will be the same of what's available via the main channel. In terms of distinguishing the advisories, that's pretty easy, as they have a different numbering format, i.e. `ALAS2<pkgname>-<year>-<number>` compared to just `ALAS2-<year>-<number>`. On the other hand, I think packages will pretty much look the same when enumerated out of the rpm database, so the best bet is likely looking at the packages' major version, and just compare with advisories that have the same, although it's a bit of a heuristic.

knqyf263:
>so the best bet is likely looking at the packages' major version, and just compare with advisories that have the same, although it's a bit of a heuristic.

For example, Golang has the same major version for standard and extras packages, right?

knqyf263:
Ideally, we'd want to identify the extras packages somehow, but we don't know of a good way to do it yet.

piax93:
I'll close this as it's pretty much lost in the sands of time.

PR #272, opened 2024-02-23
fix: revert changes on error

knqyf263:
## Description
If the update script fails midway through, the state of the directory remains intact and is accidentally committed in another step. Then, it will fail to build trivy-db. Failure of vuln-list-update is expected and must not affect trivy-db builds. The changes should be reverted in case of an error.

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -1,4 +1,13 @@
-#!/bin/bash -eu
+#!/bin/bash -eEu
+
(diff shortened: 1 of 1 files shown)


knqyf263:
Will merge this PR as it is a hotfix. The cronjob will still fail due to the change of the repository name regarding CBL-Mariner, but I'd confirm this PR correctly reverts changes.

PR #294, opened 2024-06-05
feat(azure): add support for azurelinux OVAL

tofay:
As part of https://github.com/aquasecurity/trivy/issues/6673, add support for reading the azure linux 3.0 OVAL which resides alongside the cbl-mariner 1.0 and 2.0 OVAL.

I thought it sensible to use the same package in this repo for mariner and azure linux since the OVAL parsing is the same for both.


knqyf263:
Thanks for your contribution! While it looks good, we also need to update [trivy-db](https://github.com/aquasecurity/trivy-db/tree/79d0fbd1e246f3c77eef4b9826fe4bf65940b221/pkg/vulnsrc/mariner) before merging this PR. Otherwise, it also inserts Azure Linux 3.0 into trivy-db as CBL-Mariner 3.0.

tofay:
It sounds like separating mariner and azure info in the vuln-list directory will make compatibility between the trivy repos simpler, and potentially allow easier removal of CBL-Mariner in future when 2.0 is EOL?

tofay:
I've done that separation now, but happy to revert and update trivy-db to rename mariner/3.0 to azurelinux if you prefer that approach.

knqyf263:
> It sounds like separating mariner and azure info in the vuln-list directory will make compatibility between the trivy repos simpler, and potentially allow easier removal of CBL-Mariner in future when 2.0 is EOL?

Sounds like a plan.

mariner/mariner.go
```diff
@@ -76,9 +78,13 @@ func (c Config) Update() error {
...
-	if err := os.RemoveAll(c.dir); err != nil {
+	if err := os.RemoveAll(filepath.Join(c.dir, cblDir)); err != nil {
 		return xerrors.Errorf("failed to remove CBL-Mariner Vulnerability Data directory: %w", err)
 	}
+	log.Printf("Remove Azure Linux Vulnerability Data directory %sn", c.dir)
+	if err := os.RemoveAll(filepath.Join(c.dir, azureDir)); err != nil {
+		return xerrors.Errorf("failed to remove Azure Linux Vulnerability Data directory: %w", err)
+	}
 
 	log.Print("Fetching CBL-Mariner Vulnerability Data")
```
DmitriyLewen, line 89:
```suggestion
	log.Print("Fetching Azure Linux and CBL-Mariner Vulnerability Data")
```

mariner/mariner.go
```diff
@@ -76,9 +78,13 @@ func (c Config) Update() error {
 	ctx := context.Background()
 
 	log.Printf("Remove CBL-Mariner Vulnerability Data directory %sn", c.dir)
```
DmitriyLewen, line 80:
```suggestion
	log.Printf("Remove CBL-Mariner Vulnerability Data directory %s", c.dir)
```

mariner/mariner.go
```diff
@@ -76,9 +78,13 @@ func (c Config) Update() error {
...
-	if err := os.RemoveAll(c.dir); err != nil {
+	if err := os.RemoveAll(filepath.Join(c.dir, cblDir)); err != nil {
 		return xerrors.Errorf("failed to remove CBL-Mariner Vulnerability Data directory: %w", err)
 	}
+	log.Printf("Remove Azure Linux Vulnerability Data directory %sn", c.dir)
```
DmitriyLewen, line 84:
```suggestion
	log.Printf("Remove Azure Linux Vulnerability Data directory %s", c.dir)
```

mariner/mariner.go
```diff
@@ -97,22 +103,22 @@ func (c Config) Update() error {
 			continue
 		}
 
-		if !strings.HasPrefix(entry.Name(), "cbl-mariner-") {
+		if !(strings.HasPrefix(entry.Name(), "cbl-mariner-") || strings.HasPrefix(entry.Name(), "azurelinux-")) {
```
DmitriyLewen, line 106: We use `cbl-mariner-` and `azurelinux-` prefixes a few times.
What if we move them to constants?

mariner/mariner.go
```diff
@@ -122,7 +128,12 @@ func (c Config) update(version, path string) error {
 	if err := xml.NewDecoder(f).Decode(&oval); err != nil {
 		return xerrors.Errorf("failed to decode xml: %w", err)
 	}
-	dirPath := filepath.Join(c.dir, version)
+	var dirPath string
+	if isAzureLinux {
+		dirPath = filepath.Join(c.dir, azureDir, version)
+	} else {
+		dirPath = filepath.Join(c.dir, cblDir, version)
+	}
```
DmitriyLewen, line 136: nit:
```suggestion
dirPath := filepath.Join(c.dir, azureDir, version)
	if !isAzureLinux {
		dirPath = filepath.Join(c.dir, cblDir, version)
	}
```

DmitriyLewen (commented):
Hello @tofay 
Thanks for your work.

I left some comments on the refactoring.

I also think we can start using `Azure Linux` in GH action:
https://github.com/aquasecurity/vuln-list-update/blob/80e370db109b720d2d1b7770e4b51db91f2cfc79/.github/workflows/update.yml#L86-L88

mariner/mariner.go
tofay: done

DmitriyLewen (approved):
LGTM

@knqyf263 take a look, when you have time, please.

knqyf263:
Sorry to be late. We're now targeting this support for v0.54.0. I'll review it shortly.
https://github.com/aquasecurity/trivy/issues/6673

knqyf263 (approved):

PR #298, opened 2024-07-11
feat: collect kubernetes cves data from kubernetes-sigs/cve-feed-osv repo

chen-keinan:
## Description

The purpose of this change is to collect kubernetes cves data from kubernetes-sigs/cve-feed-osv which is  relable and has osv format, instead of the existing k8s collector

- Related https://github.com/kubernetes-sigs/cve-feed-osv/issues/1


knqyf263:
We directly consume OSV feeds in trivy-db if it's hosted on GitHub.
https://github.com/aquasecurity/trivy-db/blob/e94dc822c86dd859b36554672b49ac91121aad75/Makefile#L70-L72

The main purpose of vuln-list is to display the advisories, provided by each vendor such as Red Hat in tar.gz or bzip2 formats,  on GitHub in an easy-to-read format and to track change history. `kubernetes-sigs/cve-feed-osv` is already supposed to fulfill this role, so I think we don't need to copy it into vuln-list anymore.

chen-keinan:
> We directly consume OSV feeds in trivy-db if it's hosted on GitHub. https://github.com/aquasecurity/trivy-db/blob/e94dc822c86dd859b36554672b49ac91121aad75/Makefile#L70-L72
> 
> The main purpose of vuln-list is to display the advisories, provided by each vendor such as Red Hat in tar.gz or bzip2 formats, on GitHub in an easy-to-read format and to track change history. `kubernetes-sigs/cve-feed-osv` is already supposed to fulfill this role, so I think we don't need to copy it into vuln-list anymore.

ok , thanks for update, I'll raise a PR for trivy-db

PR #328, opened 2025-01-30
fix(nvd): don't update last_update if it returned no records

### LINKED ISSUE #327: NVD Last Modified Date is not getting updated
Hi @knqyf263 ,

We saw that many CVE's which are modified recently by NVD are not getting updated in vuln-list-nvd.
eg.

1. CVE-2024-0727

    As per [NVD API](https://nvd.nist.gov/vuln/detail/CVE-2024-0727)
    Last modified on : 11/21/2024
    
    In [vuln-list-nvd](https://github.com/aquasecurity/vuln-list-nvd/blob/main/api/2024/CVE-2024-0727.json)
    Last modified on: 2024-10-14
    
    

2. CVE-2024- 5732

      As per [NVD API](https://nvd.nist.gov/vuln/detail/CVE-2024-5732)
      Last modified on : 11/21/2024
      
      In [vuln-list-nvd](https://github.com/aquasecurity/vuln-list-nvd/blob/main/api/2024/CVE-2024-5732.json)
      Last modified on: 24-06-11
      
There could be a lot more.

doing a ollback to last_updated.json will fix this ?
like done here. https://github.com/aquasecurity/vuln-list-nvd/pull/1

DmitriyLewen:
## Description
I found cases when `NVD` returns empty `vulnerabilities`:
- https://github.com/aquasecurity/vuln-list-update/actions/runs/11952931174/job/33319733712
- https://github.com/aquasecurity/vuln-list-update/actions/runs/11959096497/job/33340124175

For these runs we updated `last_update` time:
- https://github.com/aquasecurity/vuln-list-nvd/commit/30043e9c29cf127adf642f24d63d87b6791130e4
- https://github.com/aquasecurity/vuln-list-nvd/commit/f6c474ff1e0f0385d4e4fdeac15a2ccca6e98894

I think it was a bug in NVD because now I see vulnerabilities:
```bash
curl -s 'https://services.nvd.nist.gov/rest/json/cves/2.0/?lastModStartDate=2024-11-21T12:12:38&lastModEndDate=2024-11-21T18:09:55' | jq '.vulnerabilities | length'
208
```

But in any case, we lost changes in CVE from these intervals in`vuln-list-nvd` (see #327)

So we don't need to update `last_update` if NVD returns empty vulns.

## Related Issues
- Close #327

### DIFF
--- a/nvd/nvd.go
+++ b/nvd/nvd.go
@@ -105,11 +105,12 @@ func (u Updater) Update() error {
 
 	// If NVD didn't return records for all intervals
 	// we shouldn't update the last update as it might be a bug.
-	if totalEntries > 0 {
-		// Update last_updated.json at the end.
-		if err = utils.SetLastUpdatedDate(apiDir, u.lastModEndDate); err != nil {
-			return xerrors.Errorf("unable to update last_updated.json file: %w", err)
-		}
+	if totalEntries == 0 {
+		return nil
+	}
+	// Update last_updated.json at the end.
+	if err = utils.SetLastUpdatedDate(apiDir, u.lastModEndDate); err != nil {
+		return xerrors.Errorf("unable to update last_updated.json file: %w", err)
 	}
 
 	return nil
--- a/nvd/nvd.go
+++ b/nvd/nvd.go
@@ -89,6 +89,7 @@ func (u Updater) Update() error {
 		return xerrors.Errorf("unable to build time intervals: %w", err)
 	}
 
+	var totalEntries int
 	for _, interval := range intervals {
 		slog.Info("Fetching NVD entries...", slog.String("start", interval.LastModStartDate),
 			slog.String("end", interval.LastModEndDate))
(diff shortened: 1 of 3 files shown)


nvd/nvd.go
```diff
@@ -99,11 +100,16 @@ func (u Updater) Update() error {
...
 		}
+		totalEntries += totalResults
 	}
 
-	// Update last_updated.json at the end.
-	if err = utils.SetLastUpdatedDate(apiDir, u.lastModEndDate); err != nil {
-		return xerrors.Errorf("unable to update last_updated.json file: %w", err)
+	// If NVD didn't return records for all intervals
+	// we shouldn't update the last update as it might be a bug.
+	if totalEntries > 0 {
```
DmitriyLewen, line 108: When I was writing the PR description, I thought we should put the error back here so we could see it right away. But maybe that's too mach.

knqyf263:
Empty cases can be easily detected, but not if missing data is returned. However, there is nothing we can do about API bugs, so I think the current approach of dealing only with empty cases is fine.

nvd/nvd.go
```diff
@@ -99,11 +100,16 @@ func (u Updater) Update() error {
...
+		// Update last_updated.json at the end.
+		if err = utils.SetLastUpdatedDate(apiDir, u.lastModEndDate); err != nil {
+			return xerrors.Errorf("unable to update last_updated.json file: %w", err)
+		}
```
knqyf263, line 112: nit: I prefer the following style.

>Don’t hide happy path logic inside a nest of indented braces

https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

```suggestion
	if totalEntries == 0 {
		return nil
	}
```

DmitriyLewen: fixed in https://github.com/aquasecurity/vuln-list-update/pull/328/commits/86d43fe7492455d9130475e2da8aa3a1f63bb3c4

knqyf263 (approved):

PR #432, opened 2026-04-10
ci: switch to ubuntu-2404-4core runner for vuln-list update workflows

DmitriyLewen:
## Summary
- Bump runner from `ubuntu-2404-2core` to `ubuntu-2404-4core` in `update.yml`
- The 2-core runner runs out of resources during vuln-list updates, causing job failures

## Related failures
- https://github.com/aquasecurity/vuln-list-update/actions/runs/24176387674/job/70558685207

### DIFF
--- a/.github/workflows/redhat.yml
+++ b/.github/workflows/redhat.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   update:
     name: Update vuln-list-redhat
-    runs-on: ubuntu-2404-2core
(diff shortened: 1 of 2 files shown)


knqyf263:
Both Red Hat and SUSE CVRF issues have been fixed in #435 and #437 respectively. Let's try running on 2-core runners first and upgrade to 4-core only if needed.

PR #436, opened 2026-04-13
Download echo OSV advisories to vuln-list

orizerah:
Add support for downloading Echo's OSV advisory

Issue: https://github.com/aquasecurity/trivy/discussions/10519


echo/testdata/osv-golden/echo-osv/openssh/ECHO-003f-2632-599c.json
DmitriyLewen, line 1: Can you use real advisories (names, fixed version, etc) in test?

echo/testdata/osv-golden/echo-osv/openssh/ECHO-003f-2632-599c.json
```diff
@@ -0,0 +1,31 @@
+{
+  "id": "ECHO-0001-aaaa-bbbb",
+  "modified": "2024-06-15T12:00:00Z",
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "Echo",
```
DmitriyLewen, line 10: What about ecosystem for language packages?

I found ECHO-7db2-03aa-5591 with `"ecosystem":"PyPI","name":"pip",`.
Do you want use default ecosystem for language packages?

DmitriyLewen: can you also add testcase with multiple affected packages.

echo/osv_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+			testDir := t.TempDir()
+			testURL := ts.URL + "/osv/all.zip"
+			if tt.path != "" {
+				testURL = ts.URL + tt.path
+			}
+
+			ecosystems := map[string]osv.Ecosystem{
+				"echo": {
+					Dir: "",
```
DmitriyLewen: It's better to use the original folder here to verify the file paths.
```suggestion
					Dir: "echo-osv",
```

orizerah:
Hey @DmitriyLewen, I addressed you comments :)

DmitriyLewen:
Hello @orizerah 
Thanks!

what about [this comment](https://github.com/aquasecurity/vuln-list-update/pull/436#discussion_r3078330516)?

Are you planning to expand your ecosystem?
Or will you use the default ecosystems for language packages?

echo/testdata/osv-golden/echo-osv/openssh/ECHO-003f-2632-599c.json
orizerah: I'm not sure I fully understand what you mean in the question. We follow the pep440 conventions of versioning (using local versions for pypi packages).

DmitriyLewen: No, I was talking about the [affected[].package.ecosystem](https://ossf.github.io/osv-schema/#affectedpackage-field) field.
You have a specific Echo ecosystem and use Echo for deb packages.
e.g.:
```json
      "package": {
        "ecosystem": "Echo",
        "name": "python-pip"
      },
```
But i see that you use "default" ecosystems for language packages
e.g.:
```json
      "package": {
        "ecosystem": "PyPI",
        "name": "pip",
        "purl": "pkg:pypi/pip"
      },
```

Since you have your "own" ecosystem — using "default" ecosystems is a bit confusing.
That's why I had the question — is this expected, and do you plan to "expand" the Echo ecosystem in the future (for example "Echo PyPI" or something along those lines)?

orizerah: So we publish advisories both for OS packages and for language packages in OSV.
I assumed that the default ecosystem is relative to the package ecosystem, is it not how it works? For "Echo" packages it should be our versioning logic, and for PyPi I assumed that "default" means pep440.

orizerah:
We have issues with the OSV records, so I'm drafting this PR until we fix it

echo/testdata/osv-golden/echo-osv/openssh/ECHO-003f-2632-599c.json
orizerah: Hey @DmitriyLewen, we changed our osv pypi records to have a new ecosystem `Echo:PyPi`

orizerah:
@DmitriyLewen we fixed the issue with OSV and I made some changes to the code.

main.go
```diff
@@ -40,7 +40,7 @@ import (
 var (
 	target = flag.String("target", "", "update target (nvd, alpine, alpine-unfixed, redhat, redhat-oval, "+
 		"redhat-csaf-vex, debian, ubuntu, amazon, oracle-oval, suse-cvrf, photon, arch-linux, glad, cwe, osvdev, mariner, kevc, wolfi, "+
-		"chainguard, azure, openeuler, echo, minimos, eoldates, rootio)")
+		"chainguard, azure, openeuler, echo, echo-osv, minimos, eoldates, rootio)")
```
DmitriyLewen, line 44: We will merge this and related PRs together.
So should we save previous source (I mean `echo` dir)?

echo/osv.go
```diff
@@ -0,0 +1,50 @@
+package echo
+
+import (
+	"github.com/aquasecurity/vuln-list-update/osv"
+	"github.com/aquasecurity/vuln-list-update/utils"
+)
+
+const (
+	osvURL = "https://advisory.echohq.com/osv/all.zip"
+	osvDir = "echo-osv"
```
DmitriyLewen, line 10: same question.
`osv` package removes previous files, so we can use `echo` dir here.

main.go
orizerah: We need the echo dir for the os package scanning, but the OSV mostly for our Language Package advisories, should we filter them to decrease the dir size?

DmitriyLewen: So the OSV package doesn't contain advisories for OS packages?
Or is the list incomplete?
I just saw a file for OS packages in the tests and thought the OSV archive contained all your advisories.

orizerah: It contain all of our advisories. But we still want Trivy to work with our `data.json` file for advisories.

DmitriyLewen: Having 2 sources complicates the code and maintenance:
- we need to keep track of 2 packages
- the code must exclude advisories (e.g. OS packages from OSV and language packages (if any) from data.json)

What positive aspects or necessary points do you see for using 2 databases?
Perhaps we can address them right away.

orizerah: We build our OSV and `data.json` advisories differently. We rely on our findings from scanning our images with Trivy (using the `data.json`) to build the OSV advisory for other scanners.

Language packages are different, and we were asked to implement it using OSV to remain aligned with other vendors.

The `data.json` doesn't hold language packages (so no change is necessary), but I will filter out the OS packages from `echo-osv`.

DmitriyLewen: > We build our OSV and data.json advisories differently. We rely on our findings from scanning our images with Trivy (using the data.json) to build the OSV advisory for other scanners.

I compared the OSV source and the data.json file and found that the OSV feed does not contain all the advisories for OS packages (compared to data.json).
If I understood correctly — this is exactly what you meant.


Could you clarify how feasible it is to "populate" the OSV source with all the advisories?

In the meantime we will discuss the option of using 2 sources internally within the team.

orizerah: Hey @DmitriyLewen 
It's currently not very feasible to populate the OSV with the full `data.json` data. That's because we build it using the results we get from scanning with the `data.json` itself.

I changed the code here to only import the language packages advisories from OSV so it will reduce the amount of data we add to the db.

DmitriyLewen: > I changed the code here to only import the language packages advisories from OSV so it will reduce the amount of data we add to the db.

Hello @orizerah 
Thanks!

As I mentioned earlier, I noticed that the OSV feed contains only a portion of the advisories.
Could you share information on why you only store a portion of them?
And is it possible to keep only the advisories for language packages in the OSV feed (so that the filtering is on your side)?

orizerah: Hey @DmitriyLewen 

The OSV feed contains only a portion of the CVEs because we build it based on CVEs we detect using other advisory sources. To be concise, we detect CVEs using our `data.json` as a source (which is built to contain a larger scope of CVEs), and those detections are used to feed OSV.

Unfortunately, we can't keep only language package advisories in OSV because other scanners use it to scan our images with it.

knqyf263: Given the coverage difference, we need to keep both sources for now: `echo` for OS packages and `echo-osv` for language packages after filtering out the OS entries. We plan to move the Echo data to a separate `vuln-list-echo` repository, following the existing `vuln-list-redhat` and `vuln-list-debian` split. We will handle that on our side, so no repository layout changes are needed in this PR.

.github/workflows/update.yml
```diff
@@ -236,6 +236,10 @@ jobs:
           repository: ${{ env.VULN_LIST_DIR }}
           commit-message: "Echo CVE Data"
 
+      - if: always()
+        name: Echo OSV
+        run: ./scripts/update.sh echo-osv "Echo OSV Data"
```
knqyf263, line 241: `scripts/update.sh` no longer exists on the current `main`; the updater steps have moved to the `./.github/actions/update-source` composite action. Could you migrate this step to the composite action as well, using `target: echo-osv` and `commit-message: "Echo OSV Data"`?

echo/testdata/osv-golden/echo-osv/openssh/ECHO-003f-2632-599c.json
knqyf263: You mentioned that the feed was changed to `Echo:PyPi`, but the canonical OSV ecosystem name is `PyPI`, and OSV.dev's Echo ecosystem handling uses `Echo:PyPI`. The current Echo feed still publishes `Echo:PyPi`. Could you update the upstream feed to use `Echo:PyPI`? The fixture and code comment in this PR already use the canonical spelling.

.github/workflows/update.yml
orizerah: Done

echo/testdata/osv-golden/echo-osv/openssh/ECHO-003f-2632-599c.json
orizerah: I fixed that in our end, it will take some time for it to update in OSV (I requested them to sync the data in their platform).

echo/osv_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestOSVUpdater_Update(t *testing.T) {
+	tests := []struct {
+		name         string
+		path         string
+		wantFiles    []string
+		notWantFiles []string
+		wantErr      string
+	}{
+		{
+			name: "happy path",
```
DmitriyLewen, line 26: nit:
The feed contains advisories that have both an OS package (`Echo`) and an application package (`Echo:*`): 76 with `Echo:PyPI`, 4 with `Echo:Maven`, and 1 with `Echo:npm`.
In 35 of them `Echo` comes first in `affected`, so the directory is chosen after filtering (`Affected[0]`), and partial filtering (`parsed.Affected[:0]` in `osv/osv.go`) is actually exercised.

Currently every advisory in `testdata/osv-all.zip` has a single ecosystem, so this scenario is not covered.
Could you add an advisory to the test archive where `Echo` comes first followed by `Echo:PyPI`, along with a golden file for it, to verify that the OS entry is dropped and the file is written under the application package directory?

echo/testdata/osv-golden/echo-osv/pip/ECHO-7db2-03aa-5591.json
```diff
@@ -0,0 +1,38 @@
+{
+  "id": "ECHO-7db2-03aa-5591",
+  "modified": "2026-03-26T11:54:03.165Z",
+  "upstream": [
+    "CVE-2026-1703"
+  ],
+  "aliases": [
```
DmitriyLewen, line 7: nit:
This file has `aliases` populated, but in the current feed this field is empty for all application advisories (588 out of 588): CVE/GHSA IDs come only in `upstream`.
Could you update the test data to match the current feed format (no `aliases`, IDs only in `upstream`), so the test covers what is actually written to vuln-list?

osv/types.go
```diff
@@ -3,6 +3,11 @@ package osv
 type Ecosystem struct {
 	Dir string
 	URL string
+	// Filter is an optional predicate to drop individual Affected entries.
+	// If it returns true for a given Affected, that entry is excluded from
+	// the written output. Files that end up with no remaining Affected
+	// entries after filtering are skipped entirely.
+	Filter func(Affected) bool
```
DmitriyLewen, line 10: From the name `Filter` alone it's unclear whether `true` means keep or drop the entry; you only learn it from the comment.
What about a more explicit name, e.g. `Exclude` (or `Skip`)?

```go
// Exclude is an optional predicate to drop individual Affected entries.
Exclude func(Affected) bool
```

Then `Exclude: IsOSPackage` reads unambiguously.

DmitriyLewen:
I found files that appear in `all.zip` multiple times under the same name but with different content: 32 IDs occur 2–3 times.

For example, `ECHO-8ffe-c246-e62c.json` (`Echo:npm/path-to-regexp`) is in the archive three times:

```bash
$ unzip -p all.zip ECHO-8ffe-c246-e62c.json | jq -c '{modified, withdrawn, events: [.affected[].ranges[].events[]]}'
{"modified":"2026-05-07T14:20:44.772Z","withdrawn":null,"events":[{"introduced":"0"},{"fixed":"0.1.10+echo.1"}]}
{"modified":"2026-07-01T11:55:37.448Z","withdrawn":"2026-07-01T11:55:37.448Z","events":[{"introduced":"0"}]}
{"modified":"2026-05-07T14:20:44.772Z","withdrawn":"2026-05-07T14:20:44.772Z","events":[{"introduced":"0"},{"fixed":"0.1.10+echo.1"}]}
```

During extraction each subsequent copy overwrites the previous one, so vuln-list gets the last file in the archive — in this example the third copy, which is withdrawn and has an older `modified` than the second one.

Is this expected?

orizerah:
We fixed all of the comment, your last comment was actually an exporting issue in our part, so we fixed it internally.

DmitriyLewen (approved):
LGTM

DmitriyLewen:
Hello @orizerah 
Thanks for the fixes, the code looks good.
I ran `echo-osv` against the current feed (downloaded on 2026-09-24) and checked the generated files: 641 advisories, 626 of them not withdrawn.
The code stores the data correctly, but the feed itself has issues that will affect Trivy results once aquasecurity/trivy-db#659 and aquasecurity/trivy#10555 land.

| # | Issue | Example | Count (not withdrawn) | Impact on Trivy |
|---|---|---|---|---|
| 1 | `fixed` in `Echo:PyPI` is not a valid PEP 440 version (Debian versions with an epoch, Go pseudo-versions, etc.) | `ECHO-1b5a-ee63-2c52`: `django` `fixed: 3:5.2.16-1`<br>`ECHO-7d6b-5585-901d`: `paramiko` `fixed: 0.0.0-20231218163308-9d2ee975ef9f`<br>`ECHO-4cb4-2781-3bb1`: `cryptography` `fixed: 1.1.1w-0+deb11u2` | 30 (`django`, `paramiko`, `cryptography`, `webob`, `xmltodict`, `azure-identity`) | The constraint fails to parse (`improper constraint: <3:5.2.16-1`), so Trivy logs a warning and treats the package as not vulnerable. The vulnerability is not detected (false negative). |
| 2 | `fixed` in `Echo:PyPI` is a Debian version that happens to parse as PEP 440 but means something else | `ECHO-0555-db3f-4098`: `aiohttp` `fixed: 3.7.4-1+deb11u2` (parsed as `3.7.4.post1+deb11u2`) | 80 | The range becomes `<3.7.4.post1+deb11u2`, so e.g. `aiohttp 3.13.2+echo.1` is not reported as vulnerable (false negative). Users see the Debian version in the Fixed Version column. |
| 3 | The same package is listed twice with a different name case and different `fixed` values | `ECHO-15a8-1150-dd0d`: `mako` `fixed: 1.3.10+echo.1` and `Mako` `fixed: 1.3.12` (both `pkg:pypi/mako`) | 37 (`fixed` differs in 10 of them) | trivy-db normalizes pip names and merges the ranges: `<1.3.10+echo.1 \|\| <1.3.12`. The patched build `1.3.10+echo.1` passes only because it exactly matches a patched version, while `1.3.10+echo.2` is reported as vulnerable (false positive). |
| 4 | Non-standard `+e2` suffix instead of `+echo.N` | `ECHO-abd9-179e-e8ab`: `authlib` `fixed: 1.6.6+e2` | 1 | `1.6.6+e2` sorts below `1.6.6+echo.1`, so `1.6.6+echo.2` is reported as vulnerable (false positive, if `+echo.2` was meant). |
| 5 | Maven package without a groupId | `ECHO-97f7-6f4b-017b`: `commons-beanutils`, `purl: pkg:maven/commons-beanutils` (a correct `commons-beanutils:commons-beanutils` entry is next to it) | 3 | No impact yet, since trivy-db#659 supports PyPI only. Once Maven is supported, this entry never matches a package (Trivy uses `groupId:artifactId` names), so it is just noise in the DB. |

One more consequence of item 3: vuln-list gets both `echo-osv/Jinja2/` and `echo-osv/jinja2/`, as well as `echo-osv/Werkzeug/` and `echo-osv/werkzeug/`.
On case-insensitive file systems (macOS, Windows) they collide when vuln-list is cloned.

Items 1 and 2 look like OS package versions that ended up in the `Echo:PyPI` ecosystem.
Is this a delay in the feed update?

PR #437, opened 2026-04-15
fix(suse): download CVRF archive instead of individual files

knqyf263:
## Summary
- Replace 33,772 individual HTTP requests with a single `cvrf.tar.bz2` archive download
- Stream-process the archive using `compress/bzip2` + `archive/tar` (one entry at a time)
- Add 30-minute HTTP client timeout

The SUSE FTP server responds in ~2s per file, making individual downloads take 84 min+.
The archive download completes in ~6 min, resolving runner failures.

## Verified
- Unit tests pass
- Local run: output matches vuln-list repo (33,690 vs 33,476 files; diff is new entries since last update)
- GitHub Actions: SUSE CVRF step [passed](https://github.com/aquasecurity/vuln-list-update/actions/runs/24435899878) (previously crashed the runner)

Ref: https://github.com/aquasecurity/vuln-list-update/actions/runs/24176387674/job/70558685207


suse/cvrf/cvrf.go
```diff
@@ -1,116 +1,129 @@
...
 	"github.com/spf13/afero"
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/vuln-list-update/utils"
 )
 
 var (
-	cvrfURL     = "http://ftp.suse.com/pub/projects/security/cvrf/"
-	fileRegexp  = regexp.MustCompile(`<a href="(cvrf-(.*?)-.*)">.*`)
-	retry       = 5
```
DmitriyLewen, line 24: I got a 500 error when trying to download the archive.
So perhaps it makes sense to revert the retry logic.

I have often seen retries for the SUSE action, so archive downloading can also be unstable.

suse/cvrf/cvrf.go
```diff
@@ -1,116 +1,129 @@
...
+			continue
+		}
 
-func (c Config) update(os string, urls []string) error {
-	cvrfXmls, err := utils.FetchConcurrently(urls, concurrency, wait, retry)
-	if err != nil {
-		log.Printf("failed to fetch CVRF data from SUSE. err: %s", err)
-	}
+		filename := filepath.Base(hdr.Name)
+		if !strings.HasSuffix(filename, ".xml") {
```
DmitriyLewen, line 91: nit:

```suggestion
		// archive contains non-XML files (e.g. LICENSE), so skip them
		if !strings.HasSuffix(filename, ".xml") {
```

suse/cvrf/cvrf.go
knqyf263: Previously retries were there because a failure on any single file would fail the whole run, so individual retries made sense. With a single archive download I initially thought just letting the next scheduled run take care of it would be fine — but you're right that a transient 5xx is worth retrying inline.

Switched to `utils.FetchURL`, which gives us retry with exponential backoff. This also means giving up streaming and loading the whole archive into memory, but I think that's actually preferable here: retrying a stream that failed mid-transfer is tricky, and a mid-transfer disconnection would surface as a confusing truncated-tar error rather than a clean HTTP failure. A few hundred MB fits comfortably on CI runners, so the tradeoff seems fine. Added a comment explaining this.

Sorry, I accidentally force-pushed after your reviews. The change was basically just the switch to `utils.FetchURL`.

suse/cvrf/cvrf.go
knqyf263: Done.

DmitriyLewen (approved):
LGTM

Before merging:
these changes most likely shouldn't have any impact, but could you re-run the pipeline to make sure nothing is broken?

knqyf263:
Re-ran the pipeline on this branch: https://github.com/aquasecurity/vuln-list-update/actions/runs/24656940982

SUSE CVRF step passed. The only failures are `Root CVE Tracker` and `Microsoft Teams Notification`, which are pre-existing issues unrelated to this change.

PR #441, opened 2026-04-27
feat(oracle/oval): capture CVSS and CVE metadata

manojkrishnanomula:
## Summary

Oracle OVAL advisories can include extra attributes on each `<cve>` element: `public`, `cvss2`, and `cvss3`. These are stored **verbatim** in the emitted JSON (Oracle uses `score/vector` strings). Downstream (e.g. trivy-db) can split score vs vector.

**CVSS v4:** Oracle’s current OVAL feed does not expose `cvss4`; the struct field is intentionally omitted until Oracle publishes it in OVAL, so we do not imply V4 support or lock in a field name prematurely.

With the addition of new fields the size has changed from 150MB to 153MB only 3MB increased.


oracle/oval/types.go
```diff
@@ -24,7 +26,42 @@ type Reference struct {
...
+	CVSS2 string `xml:"cvss2,attr" json:"-"`
+	CVSS3 string `xml:"cvss3,attr" json:"-"`
+	CVSS4 string `xml:"cvss4,attr" json:"-"`
+
+	CVSS2Score  string `json:",omitempty"`
+	CVSS2Vector string `json:",omitempty"`
+	CVSS3Score  string `json:",omitempty"`
+	CVSS3Vector string `json:",omitempty"`
+	CVSS4Score  string `json:",omitempty"`
+	CVSS4Vector string `json:",omitempty"`
```
DmitriyLewen: We don't change the original data for advisories.
In this case it means that we save only the `CVSS*` fields and parse them in another place (e.g. in trivy-db).

oracle/oval/testdata/golden/ELSA-2007-0057-invalid-title.json
DmitriyLewen, line 1: can you update invalid-title-format.xml.bz2 file instead of creating new golden file?

oracle/oval/testdata/ELSA-2007-0057.xml
```diff
@@ -34,8 +34,8 @@ ELSA-2007-0057:  Moderate: bind security update  (MODERATE)
 <severity>MODERATE</severity>
 <rights>Copyright 2007 Oracle, Inc.</rights>
 <issued date="2007-06-26"/>
-<cve href="http://linux.oracle.com/cve/CVE-2007-0493.html">CVE-2007-0493</cve>
-<cve href="http://linux.oracle.com/cve/CVE-2007-0494.html">CVE-2007-0494</cve>
```
DmitriyLewen, line 38: should we remove these lines?

oracle/oval/types.go
manojkrishnanomula: ok, got it.

oracle/oval/testdata/golden/ELSA-2007-0057-invalid-title.json
manojkrishnanomula: done

oracle/oval/types.go
```diff
@@ -24,7 +24,13 @@ type Reference struct {
 type Cve struct {
 	Impact string `xml:"impact,attr"`
 	Href   string `xml:"href,attr"`
-	ID     string `xml:",chardata"`
+	Public string `xml:"public,attr" json:",omitempty"`
+	// Oracle encodes each CVSS version as "score/vector" (e.g. cvss3="7.3/CVSS:3.1/AV:N/...").
+	// Stored verbatim; downstream consumers split into score/vector themselves.
+	CVSS2 string `xml:"cvss2,attr" json:",omitempty"`
+	CVSS3 string `xml:"cvss3,attr" json:",omitempty"`
+	CVSS4 string `xml:"cvss4,attr" json:",omitempty"`
```
DmitriyLewen: I checked that the Oracle OVAL XML file does not contain CVSS V4 scores. RedHat doesn't contain them either. (Oracle uses their advisory structure.) But they have migrated to CSAF VEX, so I think they will not update the OVAL schema.
Most likely Oracle will use the cvss4 field, but we cannot be sure of it. Therefore I suggest removing this line until they add CVSS V4.
To avoid:
- expecting CVSS V4 scores/vectors.
- missing data if Oracle ends up using a different field name.

After CVSS V4 is added — we will include it in a separate PR.

DmitriyLewen (commented):
LGTM

left single comment.

Could you also write down the size changes before and after these changes in PR description?

DmitriyLewen (approved):
LGTM

PR #445, opened 2026-05-06
ci: switch from PAT to GitHub App installation token

DmitriyLewen:
## Description

Replace the long-lived PAT (`SA_GH_VULN_LIST_UPDATE_TOKEN`) with short-lived installation tokens minted from a GitHub App via [`actions/create-github-app-token`][action] (`v3.1.1`, pinned by SHA). App is authenticated via Client ID (`app-id` input is deprecated upstream, [we use `client-id`][client-id-doc]).

App credentials: `SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID` + `SA_GH_VULN_LIST_UPDATE_GH_APP_PRIVATE_KEY`.

Tokens are scoped to the single repository each workflow checks out: `vuln-list`, `vuln-list-redhat`, `vuln-list-nvd`, `vuln-list-debian`.

[action]: https://github.com/actions/create-github-app-token
[client-id-doc]: https://github.com/actions/create-github-app-token/blob/1b10c78c7865c340bc4f6099eb2f838309f1e8c3/action.yml

## Special case: Red Hat Security Data API

`Red Hat Security Data API` alone takes ~1.5h — longer than the 1h installation token TTL. The redhat job is restructured so this source runs last, and is split into:

1. **Fetch** — runs `vuln-list-update`; sets `has_changes=true` if there's anything to push.
2. **Refresh token** — generates a fresh installation token; skipped if no changes.
3. **Push** — drops the now-stale extraheader written by `actions/checkout`, sets `origin` URL with the refreshed token, then pushes.

Red Hat OVALv2 (~4 min) and Red Hat CSAF VEX (~1 min) keep using the initial job-level token — both fit within its 1h TTL.

## Required admin actions

- [x] **Add secrets:** `SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID`, `SA_GH_VULN_LIST_UPDATE_GH_APP_PRIVATE_KEY`.
- [ ] **Remove (after merge):** `SA_GH_VULN_LIST_UPDATE_TOKEN`.

## Test runs

Verified end-to-end on a fork (`DmitriyLewen/vuln-list-update`) using `ubuntu-24.04` runner, pointing to forked `vuln-list-*` repos.

| Workflow | Scenario | Status | Run |
|---|---|---|---|
| nvd | happy path | ✅ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/25546309907 |
| debian | happy path | ✅ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/25546311196 |
| update (all 17 sources) | happy path | ⚠️ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/25546312301 |
| redhat (full ~2h) | happy path: OVALv2 → CSAF VEX → Security Data API split | ✅ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/25547930623 |
| redhat | stub: fetch succeeds with no workdir changes → refresh + push skipped | ✅ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/25546319814 |
| redhat | stub: `vuln-list-update` exits 1 → refresh + push skipped (failure is the test) | ❌ (expected) | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/25547167304 |

> ⚠️ `update.yml` failure is unrelated to this migration: only the `openEuler CVE` step failed with `Timeout Fetching URL` from `repo.openEuler.org` (transient external API). All 16 other sources, the App-token generation, and the pushes succeeded.

### DIFF
--- a/.github/workflows/debian.yml
+++ b/.github/workflows/debian.yml
@@ -27,7 +27,7 @@ jobs:
       id: vuln-list-debian-token
       uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
       with:
-        app-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID }}
+        client-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID }}
         private-key: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_PRIVATE_KEY }}
         owner: ${{ github.repository_owner }}
         repositories: ${{ env.VULN_LIST_DIR }}
--- a/.github/workflows/debian.yml
+++ b/.github/workflows/debian.yml
@@ -27,7 +27,7 @@ jobs:
       id: vuln-list-debian-token
       uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
       with:
-        app-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_ID }}
+        app-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID }}
         private-key: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_PRIVATE_KEY }}
         owner: ${{ github.repository_owner }}
         repositories: ${{ env.VULN_LIST_DIR }}
--- a/.github/workflows/debian.yml
+++ b/.github/workflows/debian.yml
@@ -23,11 +23,20 @@ jobs:
       with:
         go-version-file: go.mod
 
+    - name: Generate token for vuln-list-debian
+      id: vuln-list-debian-token
+      uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+      with:
+        app-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_ID }}
+        private-key: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_PRIVATE_KEY }}
+        owner: ${{ github.repository_owner }}
+        repositories: ${{ env.VULN_LIST_DIR }}
+
     - name: Check out vuln-list-debian repo
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       with:
         repository: ${{ github.repository_owner }}/${{ env.VULN_LIST_DIR }}
-        token: ${{ secrets.SA_GH_VULN_LIST_UPDATE_TOKEN }}
+        token: ${{ steps.vuln-list-debian-token.outputs.token }}
         path: ${{ env.VULN_LIST_DIR }}
         persist-credentials: true # update.sh runs git push
 
--- a/.github/workflows/nvd.yml
+++ b/.github/workflows/nvd.yml
@@ -27,7 +27,7 @@ jobs:
       id: vuln-list-nvd-token
       uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
       with:
-        app-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID }}
+        client-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID }}
         private-key: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_PRIVATE_KEY }}
         owner: ${{ github.repository_owner }}
         repositories: ${{ env.VULN_LIST_DIR }}
--- a/.github/workflows/nvd.yml
+++ b/.github/workflows/nvd.yml
@@ -23,11 +23,20 @@ jobs:
       with:
         go-version-file: go.mod
 
+    - name: Generate token for vuln-list-nvd
+      id: vuln-list-nvd-token
+      uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+      with:
+        app-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_ID }}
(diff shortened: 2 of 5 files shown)


knqyf263 (approved):
The redhat split is a clever workaround for the 1h TTL, but the same cumulative-time issue can hit `update.yml` too (17 sources sharing one job-level token). Rather than special-casing redhat, I think we can unify all four workflows on a single pattern. Two ideas combined:

1. **Check out the target repo without the App token.** All `vuln-list-*` repos are public, so `actions/checkout` does not need the App token. Default GITHUB_TOKEN with `persist-credentials: false` is enough, and it also avoids leaking auth state into the workdir.

2. **Generate a fresh token right before each push, and revoke it right after.** Call `actions/create-github-app-token` immediately before `git push`, then `gh api -X DELETE /installation/token` immediately after. Each token only lives for a few seconds, so fetch duration no longer matters.

To avoid repeating this five-step sequence (fetch, commit, generate token, push, revoke) in every workflow, we can move it into a single composite action at `.github/actions/run-vuln-list-source/action.yml`. **This composite action would replace `scripts/update.sh` entirely.** Sketch:

`.github/actions/run-vuln-list-source/action.yml`:

```yaml
name: "Run vuln-list source"
description: "Fetch one source, commit, push with a fresh App token, then revoke it"
inputs:
  target:        { required: true }
  message:       { required: true }
  vuln-list-dir: { required: true }
  client-id:     { required: true }
  private-key:   { required: true }
runs:
  using: composite
  steps:
    - name: Fetch
      shell: bash
      env:
        TARGET: ${{ inputs.target }}
        VULN_LIST_DIR: ${{ inputs.vuln-list-dir }}
      run: |
        ./vuln-list-update -vuln-list-dir "$VULN_LIST_DIR" -target "$TARGET" || {
          echo "[Err] Revert changes" >&2
          cd "$VULN_LIST_DIR" && git reset --hard HEAD
          exit 1
        }

    - name: Stage changes
      id: stage
      shell: bash
      working-directory: ${{ inputs.vuln-list-dir }}
      env:
        MESSAGE: ${{ inputs.message }}
      run: |
        if [[ -n $(git status --porcelain) ]]; then
          git add .
          git commit -m "$MESSAGE"
          echo "has_changes=true" >> "$GITHUB_OUTPUT"
        fi

    - name: Generate token
      if: steps.stage.outputs.has_changes == 'true'
      id: token
      uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
      with:
        client-id:    ${{ inputs.client-id }}
        private-key:  ${{ inputs.private-key }}
        owner:        ${{ github.repository_owner }}
        repositories: ${{ inputs.vuln-list-dir }}
        skip-token-revoke: true

    - name: Push
      if: steps.stage.outputs.has_changes == 'true'
      shell: bash
      working-directory: ${{ inputs.vuln-list-dir }}
      env:
        GH_APP_TOKEN: ${{ steps.token.outputs.token }}
      run: |
        AUTH=$(printf 'x-access-token:%s' "$GH_APP_TOKEN" | base64 -w0)
        git -c "http.https://github.com/.extraheader=AUTHORIZATION: basic $AUTH" \
            push origin HEAD

    - name: Revoke token
      if: always() && steps.token.outputs.token != ''
      shell: bash
      env:
        GH_TOKEN: ${{ steps.token.outputs.token }}
      run: gh api -X DELETE /installation/token
```

Workflow side (example, `update.yml`):

```yaml
- uses: actions/checkout@SHA
  with:
    repository: ${{ github.repository_owner }}/vuln-list
    path: vuln-list
    persist-credentials: false
- uses: actions/setup-go@SHA
  with:
    go-version-file: go.mod
- run: go build -o vuln-list-update .

- if: always()
  name: Alpine Issue Tracker
  uses: ./.github/actions/run-vuln-list-source
  with:
    target: alpine
    message: "Alpine Issue Tracker"
    vuln-list-dir: vuln-list
    client-id: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_CLIENT_ID }}
    private-key: ${{ secrets.SA_GH_VULN_LIST_UPDATE_GH_APP_PRIVATE_KEY }}
# ... remaining sources
```

Effects:
- `update.yml` cumulative-TTL risk: gone
- redhat fetch/refresh/push split: gone, becomes the same shape as the others
- `scripts/update.sh`: removed, replaced by the composite action
- each token's lifetime: a few seconds (generate, push, revoke)

One caveat: we had a security incident involving a composite action in the past, so this part needs careful security review before merging.

This PR is OK to merge as-is for the PAT-to-App migration. I'm fine with handling the unification above in a follow-up PR. Let me know if you would rather fold it in here.

DmitriyLewen:
I had the same idea, but I thought about the same concerns (passing tokens into a composite action) as you did. And even if we carefully review everything now — we are not protected from mistakes in the future.
For this reason, and since we currently have a time margin for update.yaml — I decided to update the logic only for "Security Data API".

Moreover, if we have issues in update.yaml — we can simply request a new token between update script runs|steps. This will be safer, as the token will only be available within the scope of a single run/workflow.

knqyf263:
Not blocking the merged decision, but a couple of points worth recording for the trade-off:

**On the composite action risk**

I looked at the [secure-use reference](https://docs.github.com/en/actions/reference/security/secure-use) and [community discussion #34212](https://github.com/orgs/community/discussions/34212) but could not find a risk that is specific to composite actions, as opposed to risks that apply to any action handling a secret. The composite-related CVEs people cite tend to come from interpolating **untrusted event input** into a `run:` block, which is the same script-injection class we mitigate everywhere with `env:` and SHA pinning. For an internal composite invoked from `push`/`schedule` (no `pull_request_target`), I do not think passing a token in carries categorically higher risk than passing it to any other step.

**Cost of staying inlined**

1. The new fetch -> refresh -> push path for Red Hat Security Data API no longer goes through `scripts/update.sh`. Future fixes to `update.sh` silently will not apply to it.
2. `actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3` is now copy-pasted across five workflows. Rotating the pinned SHA means touching every file consistently.
3. If `update.yml` ever needs the same split, we copy-paste the 3-step block again, multiplying the surface for exactly the mistakes (extraheader, \`has_changes\`, script injection) we want to avoid.

My understanding is that avoiding composite is not inherently safer here, and being overly cautious about it may just trade one risk for another. YAGNI is fine for now, but once a second workflow needs this, I would push back on a second copy.

PR #448, opened 2026-05-19
ci: extract source fetch/commit/push into a composite action

### LINKED ISSUE #447: git push fails in update workflows after long-running fetches
## Description

`Update vuln-list repo` (and previously `Update vuln-list-redhat repo` for `Red Hat Security Data API`) fails on `git push` for any source step running more than ~1h after the job started:

```
fatal: could not read Username for 'https://github.com': No such device or address
##[error]Process completed with exit code 128.
```

Concrete example: [run 26036426564 / job 76535737088](https://github.com/aquasecurity/vuln-list-update/actions/runs/26036426564/job/76535737088). Two of 28 steps failed identically:

- step 25 `MinimOS Secdb` — failed at `T+1h 2min 20s` from token generation
- step 28 `Root CVE Tracker` — failed at `T+1h 3min 51s` from token generation

Every earlier source in the same job pushed successfully; later sources without changes silently skipped.

## Root cause

GitHub App installation tokens have a fixed 1h TTL — the limit is enforced by GitHub's API and cannot be raised.

`actions/checkout` writes the job-level token into `vuln-list/.git/config` as an `extraheader` via `persist-credentials: true`. Once the installation token expires, that header is invalid and `git push` falls back to interactive prompt — which has no stdin in CI.

Before #445, `SA_GH_VULN_LIST_UPDATE_TOKEN` was a PAT with no 1h TTL, so long jobs worked. The migration to a GitHub App in #445 introduced the limit but the workflow structure was not updated to refresh the token mid-job.

For `redhat.yml`, the same root cause was already mitigated in #445 by an inlined fetch + refresh-token + push triplet for `Red Hat Security Data API` only. The remaining workflows were not covered.

## Proposed fix

Extract the inlined fetch + refresh-token + push pattern into a reusable composite action (`.github/actions/update-source`) and apply it to every push site in `update.yml`, `redhat.yml`, `debian.yml`, and `nvd.yml`. Each call mints a fresh installation token immediately before the push, eliminating the TTL race regardless of upstream wall-clock.

The composite consolidation was suggested by @knqyf263 in [#445 (comment)](https://github.com/aquasecurity/vuln-list-update/pull/445#issuecomment-4441600758) once a second occurrence of the split appeared — this issue is that second occurrence.

DmitriyLewen:
## Related Issues
- Close #447

## Summary

This PR extracts the fetch + commit + push pattern for a single vuln-list source into a reusable composite action (`.github/actions/update-source`) and applies it to every push site across `update.yml`, `redhat.yml`, `debian.yml`, and `nvd.yml`.

The composite mints a fresh GitHub App installation token immediately before each push, so long-running workflows survive the 1h installation token TTL. This eliminates the failure mode tracked in #447:

```
fatal: could not read Username for 'https://github.com': No such device or address
```

Affected previously: `update.yml` (~1h, sporadic failures on late sources; e.g. [run 26036426564](https://github.com/aquasecurity/vuln-list-update/actions/runs/26036426564/job/76535737088)) and `redhat.yml` `Red Hat Security Data API` (~1.5h, already mitigated inline in #445).

The composite consolidation was suggested by @knqyf263 in [#445 (comment)](https://github.com/aquasecurity/vuln-list-update/pull/445#issuecomment-4441600758).

## Changes

| File | What |
|---|---|
| 🆕 `.github/actions/update-source/action.yml` | Composite: run `vuln-list-update` for the target → detect changes → mint new App token → drop extraheader → set origin URL → `git add . && git commit && git push` |
| 🗑️ `scripts/update.sh` | Removed; the fetch invocation is inlined into the composite |
| ♻️ `update.yml`, `debian.yml`, `nvd.yml`, `redhat.yml` | Each source switches from `./scripts/update.sh <target> <msg>` (or the inline fetch/refresh/push triplet in redhat) to `uses: ./.github/actions/update-source` |
| ♻️ vuln-list-* checkouts | Dropped the "Generate token for vuln-list-* checkout" step in every workflow and removed `token:` from `actions/checkout`. The target repos are public, so the default `GITHUB_TOKEN` clones them anonymously; push auth is handled by the composite via a freshly minted installation token. `persist-credentials: false` because the composite sets the origin URL directly. |

## Behaviour notes

- ~25 extra `actions/create-github-app-token` calls per `update.yml` run × ~2-3s ≈ 1-1.5min added wall-clock. Acceptable.
- `Red Hat Security Data API` no longer has to be the last step in `redhat.yml` — every push gets its own fresh token, so the ordering constraint from #445 is dropped (comment removed).
- Source order in `redhat.yml` is preserved for now.

## Test plan

Verified end-to-end on the fork.

| Workflow | Scenario | Status | Run |
|---|---|---|---|
| debian | happy path | ✅ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/26213464616 |
| nvd | happy path | ✅ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/26213465993 |
| update (all 21 sources) | 20/21 sources pushed. `openEuler CVE` failed with a `Timeout Fetching URL` panic — a pre-existing flaky upstream API documented in #447 (same step failed identically in #445's fork run for the same reason). Importantly, the composite isolated the failure: workdir was reverted, subsequent sources (echo, minimos, seal, eoldates, rootio — the previously TTL-failing tail) all pushed normally. | ⚠️ pre-existing | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/26102787562 |
| redhat (full ~2h) | happy path: OVALv2 + CSAF VEX + Security Data API; the long-running Security Data API push lands well past the 1h TTL | ✅ | https://github.com/DmitriyLewen/vuln-list-update/actions/runs/26102790004 |


knqyf263 (dismissed):
Since `scripts/fetch.sh` is now only called from this composite action, we could inline it into `action.yml` and drop the file. The script is ~10 effective lines, the `if [ -z "$TARGET" ]` check is already enforced by `target: required: true` on the composite input, and inlining keeps the whole push lifecycle in a single file.

Suggested replacement for the `Fetch` step:

```yaml
- name: Fetch
  shell: bash
  env:
    VULN_LIST_DIR: ${{ inputs.repository }}
    TARGET: ${{ inputs.target }}
  run: |
    ./vuln-list-update -vuln-list-dir "$VULN_LIST_DIR" -target "$TARGET" || {
      echo "[Err] Revert changes" >&2
      cd "$VULN_LIST_DIR" && git reset --hard HEAD
      exit 1
    }
```

Not a blocker. You can keep the script if there's a local-debug use case or something else I'm missing.

DmitriyLewen:
Personally I always test via the `./vuln-list-update ...` command, but I kept the script in case someone still uses it for local testing.
But since the only difference between the script and the `./vuln-list-update ...` command is now just one line (the script additionally does a reset) and you also thought about this — I agree with you — there is not much point in it.
Updated in https://github.com/aquasecurity/vuln-list-update/pull/448/commits/c0d3b1bde655d62a76dbceb245baad9b7414d741

knqyf263 (dismissed):
I need to correct something from my earlier review. I was under the wrong impression for a long time: `required: true` is not actually validated at runtime.

From the docs:
> Actions using `required: true` will not automatically return an error if the input is not specified.

https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax

It's fine as-is since every caller passes `target`, but we might want an explicit validation in the action rather than relying on `required: true`.

DmitriyLewen:
I'm not sure there is any point in adding a check in action.I'm not sure there is any point in adding a check in action.yaml:

- we will only see it after the PR is merged
- we will still get an error from vuln-list-update, but a slightly less informative one ("unknown target" instead of "target is required" (see https://github.com/DmitriyLewen/vuln-list-update/actions/runs/26213292073/job/77129226885)).

A better solution would be to add an additional linter (unfortunately `zizmor` doesn't catch this issue) to detect the problem at the PR stage.
But this might be excessive, because:

- IDEs usually show this error
  
	изображение

- we typically build/test changes in forks (and ask contributors to do the same)

I suggest revisiting this issue if we encounter a missed target, and for now keep the required field.

knqyf263 (approved):

PR #455, opened 2026-06-09
feat(suse): CSAF to vuln-list-suse, keep CVRF in vuln-list

manojkrishnanomula:
## Summary

Addresses review feedback from @knqyf263 and @DmitriyLewen:

- **CSAF** advisories are published to the dedicated **`vuln-list-suse`** repo (like `vuln-list-redhat`), via new `.github/workflows/suse.yml`.
- **CVRF** updater is **restored** and continues updating main `vuln-list` for legacy/premium consumers during the transition.
- CSAF data is stored **as-is** (`csaflib.Advisory` + `Validate()`), matching the `redhat/csaf` convention — no trimming in this repo.
- Shared `suse.Updater` walks CVRF/CSAF archives; CSAF skips invalid entries with logged counters (does not abort the whole archive).

### Architecture

| Target | Workflow | Destination repo | Path |
|--------|----------|------------------|------|
| `suse-cvrf` | `suse.yml` | `vuln-list` | `cvrf/suse/{suse,opensuse}/...` |
| `suse-csaf` | `suse.yml` | `vuln-list-suse` | `csaf/suse/{suse,opensuse}/...` |

### Full `csaf.tar.bz2` integrity audit (2026-08-11 archive)

| Metric | Count |
|--------|------:|
| Matched advisory JSON (`suse-su-*` / `opensuse-su-*`) | 33,853 |
| Saved | 30,821 |
| Skipped | 3,032 (9.0%) |

**Skip breakdown**

| Reason | Count |
|--------|------:|
| `json.Unmarshal` into `csaflib.Advisory` | 3,032 |
| `adv.Validate()` | 0 |
| Invalid advisory ID format | 0 |

All 3,032 unmarshal skips share a **single root cause** — an unencoded comma in OES CPE vendor names. `csaflib` rejects these three values (CPE 2.2 requires percent-encoding; correct form is `novell%2Cinc`):

```
cpe:/o:novell,inc:OES:23.4
cpe:/o:novell,inc:OES:24.4
cpe:/o:novell,inc:OES:25.4
```

OES rides along inside ordinary SLES advisories (e.g. `SUSE-SU-2024:4333-1`), so the whole document is dropped together with all its SLES products and CVEs. Every affected advisory is `suse-su` (none are `opensuse-su`). The dropped set references ~11,206 distinct CVEs.

| Year | Failing |
|------|--------:|
| 2022 | 345 |
| 2023 | 676 |
| 2024 | 347 |
| 2025 | 691 |
| 2026 | 973 |

**Urgency:** not urgent for merge. All 3,032 failing advisories are also present in the CVRF archive — while `suse-cvrf` keeps updating, there is no detection gap. The gap opens the day CVRF is retired.

**Known limitation / CVRF-retirement blocker:** CSAF cannot fully replace CVRF until SUSE fixes the OES CPE encoding. Tracking: _SUSE bug report pending — will link here once filed._

**vs CVRF (`cvrf.tar.bz2`)**

| | Count |
|--|------:|
| CVRF saved advisories | 36,586 |
| Overlapping SU advisory IDs | ~29,258 |
| CSAF-only | 1 (`openSUSE-SU-2025-15854-1`) |
| CVRF-only | 7,270 (mostly `SUSE Image SUSE-IU-*` advisories absent from CSAF) |

## Companion PR

- **trivy-db:** https://github.com/aquasecurity/trivy-db/pull/674 — adds `pkg/vulnsrc/suse-csaf` (reads `vuln-list-suse`); **`suse-cvrf` stays active in `vulnsrc.All`** until the feed gap above is resolved and pipelines are switched (similar to `redhat-csaf`).

These PRs can merge independently; OSS pipeline switch to CSAF is a follow-up.

## Test plan

- [x] `go test ./suse/csaf/...` (includes invalid-advisory skip cases)
- [x] `go test ./suse/cvrf/...`
- [x] `go test ./suse/...`
- [ ] Run `suse.yml` workflow once after merge (requires GitHub App write access to `vuln-list-suse`)
- [ ] Confirm `vuln-list-suse` is seeded before enabling CSAF in trivy-db pipelines
- [ ] File SUSE bug for OES CPE percent-encoding and link ticket here


DmitriyLewen (commented):
Two moments:
### 1. Don't trim the data — store it as-is

In `vuln-list-update` the convention is to store source data **as-is** (raw) and do trimming/normalization in downstream repos (e.g. `trivy-db`). This PR is the only place in the whole repo that manually trims data before saving (`trim`, `trimNotes`, `trimDocReferences`, `trimProductTree`, `trimVulnerabilities`, `trimScore`).

For reference:
- The sibling CSAF updater `redhat/csaf` decodes into `csaf.Advisory`, calls `advisory.Validate()`, and stores the whole structure — no trimming.
- The old SUSE CVRF updater being removed here also stored the fully parsed structure.

Downsides of trimming here:
- Any new field Trivy needs later requires changes in this repo (`trim*` + `types.go`), whereas selecting fields should be the downstream's job.
- Silent data loss by design: if SUSE adds a new field/category (e.g. CVSS v4), it just won't make it into the output.
- Because of the manual trim with hand-written nil checks, a single malformed advisory aborts the whole archive walk (`Update` returns an error from `walkArchive`), dropping every subsequent advisory for that run.

Suggestion: align this updater with `redhat/csaf` — decode into `csaflib.Advisory`, optionally `Validate()`, and store it in full (`utils.SaveCVEPerYear` / `WriteJSON`), without `trim*`. That removes the trim functions and `types.go` and resolves several issues at once.

If there's a concern about feed size (SUSE advisory-level documents carry many products/arches), let's discuss and measure it — but we shouldn't diverge from the convention by default.

### 2. A companion trivy-db PR is required

Both the output path and the format change: it used to be `cvrf/suse/...` in the CVRF schema, now it's `csaf/suse/...` in a new schema. Meanwhile `trivy-db` (`pkg/vulnsrc/suse-cvrf`) reads the old `cvrf/suse` path and decodes the CVRF structure.

Once this PR merges and CI runs, the old `cvrf/suse` data stops updating while `trivy-db` can't read the new format yet — SUSE scanning in Trivy would silently degrade.

So a corresponding `trivy-db` PR for the new path/format is needed, and both PRs must be merged **together** (or in close coordination). Please link the related trivy-db PR here.

Note that this ties back to point 1: since we store the raw CSAF here, the trimming/normalization (the logic currently in the `trim*` functions) belongs in that `trivy-db` PR, where the new format is consumed.

manojkrishnanomula:
Companion trivy-db PR is ready and linked here:\n\n- https://github.com/aquasecurity/trivy-db/pull/674 updates SUSE ingestion to read  raw advisory data (with legacy  fallback during transition).

manojkrishnanomula:
Companion trivy-db PR is ready and linked here:

- https://github.com/aquasecurity/trivy-db/pull/674

It updates SUSE ingestion to read `vuln-list/csaf/suse/...` raw advisory data (with legacy `cvrf/suse/...` fallback during transition).

knqyf263:
Two things on the transition, plus a data-integrity check.

**1. Removing the CVRF updater vs. consumers that cannot switch quickly**

For OSS this is straightforward: we merge the vuln-list-update and trivy-db PRs together, so the `cvrf/suse` fallback in trivy-db#674 is not needed. We can drop it and land both in lockstep.

The real concern is removing the `suse/cvrf` updater here. Any consumer still reading `cvrf/suse` keeps being served, but its data freezes the moment this merges. My understanding is that the premium DB pipeline reads CVRF and cannot switch to CSAF as quickly as OSS can. If that is right, removing CVRF now would stop SUSE advisory updates on that side. Is that acceptable, and is the premium side ready to move? If not, I would keep the CVRF updater running in parallel until all consumers have migrated, and remove it in a follow-up.

**2. vuln-list size / a dedicated repo**

If we keep CVRF around for the consumers above, we are now generating CVRF plus full CSAF advisories in the same repo. SUSE CSAF documents carry many products/arches, so the raw advisories are not small, and the file count grows too. vuln-list has grown large enough to crash before, so this pushes back toward that limit.

I would suggest moving SUSE into its own repo, e.g. `vuln-list-suse`, following the precedent of `vuln-list-redhat` and others. That keeps the main vuln-list from growing and gives us room to run CVRF and CSAF side by side during the transition. What do you think?

**3. Silently skipped advisories**

The updater logs and skips advisories that fail `json.Unmarshal` or `adv.Validate()`. Not aborting the whole walk is the right call, but a silent skip can quietly drop advisories, and vendor CSAF is often not fully spec-conformant. Could you run this against the full `csaf.tar.bz2` and report how many advisories are skipped vs. saved, and compare the saved count against the current CVRF advisory set? That confirms we are not dropping coverage today, even if future regressions would be hard to notice.

.github/workflows/suse.yml
```diff
@@ -0,0 +1,59 @@
...
+        git config --global user.name "GitHub Action"
+
+    - name: Compile vuln-list-update
+      run: go build -o vuln-list-update .
+
+    - if: always()
+      name: SUSE CSAF
+      uses: ./.github/actions/update-source
+      with:
+        target: suse-csaf
```
DmitriyLewen, line 56: Can we move `cvrf` from vuln-list to vuln-list-suse?

.github/workflows/suse.yml
```diff
@@ -0,0 +1,59 @@
...
+        commit-message: "SUSE CSAF"
+
+    - name: Microsoft Teams Notification
+      uses: Skitionek/notify-microsoft-teams@e7a2493ac87dad8aa7a62f079f295e54ff511d88
+      if: failure()
+      with:
+        webhook_url: ${{ secrets.TRIVY_MSTEAMS_WEBHOOK }}
+        needs: ${{ toJson(needs) }}
+        job: ${{ toJson(job) }}
+        steps: ${{ toJson(steps) }}
```
DmitriyLewen: We migrated to Slack notifications - https://github.com/aquasecurity/vuln-list-update/blob/d842f048c9efb99b64b248f8aecac3a6be589548/.github/workflows/update.yml#L285-L292

suse/csaf/csaf.go
```diff
@@ -0,0 +1,183 @@
...
+
+		if err := adv.Validate(); err != nil {
+			log.Printf("skip invalid CSAF advisory (%s): %v", e.Filename, err)
+			return nil
+		}
+
+		if adv.Document == nil || adv.Document.Tracking == nil || adv.Document.Tracking.ID == nil {
+			log.Printf("skip advisory without tracking id (%s)", e.Filename)
+			return nil
+		}
```
DmitriyLewen, line 68: Did you check the number of skipped advisories? (see point 3 in https://github.com/aquasecurity/vuln-list-update/pull/455#issuecomment-4767758437)

suse/csaf/csaf.go
```diff
@@ -0,0 +1,183 @@
...
+func decompressArchive(url string, body []byte) (io.Reader, error) {
+	switch {
+	case strings.HasSuffix(url, ".tar.bz2"):
+		return bzip2.NewReader(bytes.NewReader(body)), nil
+	case strings.HasSuffix(url, ".tar.gz"):
+		return gzip.NewReader(bytes.NewReader(body))
+	default:
+		return nil, xerrors.Errorf("unsupported archive format: %s", url)
+	}
+}
```
DmitriyLewen: IIRC `.tar.gz` is required only for tests.
Can we add a comment about this?

suse/csaf/csaf.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+}
+
+func decompressArchive(url string, body []byte) (io.Reader, error) {
+	switch {
+	case strings.HasSuffix(url, ".tar.bz2"):
+		return bzip2.NewReader(bytes.NewReader(body)), nil
+	case strings.HasSuffix(url, ".tar.gz"):
+		return gzip.NewReader(bytes.NewReader(body))
```
DmitriyLewen: You don't close reader here.

suse/csaf/csaf.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		return nil
+	})
+}
+
+func osNameFromFilename(filename string) (string, error) {
+	match := fileRegexp.FindStringSubmatch(filename)
+	if len(match) < 2 {
+		return "", fmt.Errorf("unexpected filename")
+	}
```
DmitriyLewen: You already checked this in `walkArchive`.

suse/csaf/csaf.go
```diff
@@ -0,0 +1,183 @@
...
+
+		dir := filepath.Join(csafDir, suseDir, osName)
+		if err := c.savePerYear(dir, string(*adv.Document.Tracking.ID), adv); err != nil {
+			return xerrors.Errorf("failed to save CSAF: %w", err)
+		}
+		return nil
+	})
+}
+
+func osNameFromFilename(filename string) (string, error) {
```
DmitriyLewen: I think we can remove regexp.
Something like that:
```go
  func osNameFromFilename(filename string) (osName string, ok bool) {
        switch {
        case strings.HasPrefix(filename, "suse-su-"):
                return "suse", true
        case strings.HasPrefix(filename, "opensuse-su-"):
                return "opensuse", true
        default:
                return "", false
        }
  }
```

And also move the check into the handler itself. In this case we'll be reading the file, but I assume there shouldn't be any other files in the archive. But if there are, and there are many of them, we can just do the check in the walker and pass the OS further down — or check twice.

main.go
```diff
@@ -40,7 +41,7 @@ import (
 
 var (
 	target = flag.String("target", "", "update target (nvd, alpine, alpine-unfixed, redhat, redhat-oval, "+
-		"redhat-csaf-vex, debian, ubuntu, amazon, bottlerocket, oracle-oval, suse-cvrf, photon, arch-linux, glad, cwe, osvdev, mariner, kevc, wolfi, "+
```
DmitriyLewen, line 43: don't remove bottlerocket.

README.md
```diff
@@ -20,7 +20,7 @@ https://github.com/aquasecurity/vuln-list/
 $ vuln-list-update -h
 Usage of vuln-list-update:
   -target string
-        update target (nvd, alpine, alpine-unfixed, redhat, redhat-oval, debian, ubuntu, amazon, bottlerocket, oracle-oval, suse-cvrf, photon, arch-linux, ghsa, glad, cwe, osv, mariner, kevc, wolfi, chainguard, azure, openeuler, echo, minimos, rootio, seal)
```
DmitriyLewen, line 23: same

suse/csaf/csaf.go
```diff
@@ -0,0 +1,183 @@
...
+			return nil
+		}
+
+		var adv csaflib.Advisory
+		if err := json.Unmarshal(e.Data, &adv); err != nil {
+			log.Printf("skip invalid CSAF json (%s): %v", e.Filename, err)
+			return nil
+		}
+
+		if err := adv.Validate(); err != nil {
```
DmitriyLewen, line 58: Can we add testcase for this error?

suse/csaf/csaf.go
```diff
@@ -0,0 +1,183 @@
...
+
+	yearDir := filepath.Join(c.VulnListDir, dirName, year)
+	fileName := fmt.Sprintf("%s.json", strings.Replace(advisoryID, ":", "-", 1))
+	if err := utils.WriteJSON(c.AppFs, yearDir, fileName, data); err != nil {
+		return xerrors.Errorf("failed to write file: %w", err)
+	}
+	return nil
+}
+
+func walkArchive(url string, retries int, nameRegexp *regexp.Regexp, handler func(archiveEntry) error) error {
```
DmitriyLewen: Did you create this function to unify csaf and cvrf? Let's do it. We'll move it into the suse/suse.go package and import it from there.

suse/csaf/csaf.go
manojkrishnanomula: yes, added those numbers in description.

suse/csaf/csaf.go
DmitriyLewen: I don't see tests for invalid advisories.
Did you add them?

suse/archive.go
DmitriyLewen, line 1: I refactored this package a bit to make it more readable and so that we don't have to move raw data between packages. 
https://github.com/aquasecurity/vuln-list-update/pull/455/commits/0635ccbdde20234852b91646dc42b2f9d0987763
Please check that I didn't break anything.

suse/csaf/csaf_test.go
```diff
@@ -0,0 +1,99 @@
...
+		archiveDir  string
+		goldenFiles map[string]string
+	}{
+		{
+			name:       "positive test",
+			appFs:      afero.NewMemMapFs(),
+			archiveDir: "testdata/csaf",
+			goldenFiles: map[string]string{
+				"/tmp/csaf/suse/suse/2019/SUSE-SU-2019-0048-2.json":         "testdata/golden/SUSE-SU-2019-0048-2.json",
+				"/tmp/csaf/suse/opensuse/2019/openSUSE-SU-2019-0003-1.json": "testdata/golden/openSUSE-SU-2019-0003-1.json",
```
DmitriyLewen, line 52: Can we trim down the test files? It would improve readability.

suse/csaf/csaf.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	return updater.Update(func(filename string, data []byte) (string, any, error) {
+		var adv csaflib.Advisory
+		if err := json.Unmarshal(data, &adv); err != nil {
+			log.Printf("skip invalid CSAF json (%s): %v", filename, err)
+			return "", nil, nil
+		}
+
+		if err := adv.Validate(); err != nil {
```
DmitriyLewen, line 58: Thanks for the numbers in the description — I reproduced them against a fresh archive (2026-08-11) and dug into the cause.
It turned out to be far more specific than "legacy CPE URI rejection".

**What I found**

Out of 33,853 `suse-su-*` / `opensuse-su-*` files, 3,032 (9.0%) fail to unmarshal, and not a single one fails `Validate()`.
All 3,032 share the same cause, and there are only three distinct offending values:

```
cpe:/o:novell,inc:OES:23.4
cpe:/o:novell,inc:OES:24.4
cpe:/o:novell,inc:OES:25.4
```

`csaflib` validates `cpe` against the pattern from the CSAF 2.0 schema.
The branch for the legacy CPE 2.2 URI format — `[c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\._\-~%]*){0,6}` — does not allow a comma.
And rightly so: per the CPE 2.2 spec special characters must be percent-encoded, so the correct form is `novell%2Cinc`.
In other words the SUSE feed is non-conformant and the library is behaving correctly.

**This is not old data**

| year | failing |
|---|---:|
| 2022 | 345 |
| 2023 | 676 |
| 2024 | 347 |
| 2025 | 691 |
| 2026 | 973 |

Nothing before 2022, and the largest share is from the current year.
The PR description says 2,739 — two months later it is 3,032, so the count keeps growing.

Every affected advisory is `suse-su`, none are `opensuse-su`: OES rides along inside ordinary SLES advisories.
For example `SUSE-SU-2024:4333-1` ("Security update for libaom, libyuv") — a single OES entry under `product_tree/.../product_identification_helper` drops the whole document together with all its SLES products and CVEs.
In total the dropped advisories reference 11,206 distinct CVEs.

**How urgent is it**

Not urgent.
I matched the filenames against the CVRF archive: all 3,032 are present there as well.
While `suse-cvrf` is alive there is no detection gap — the same advisories arrive in CVRF form.
The gap opens the day CVRF is retired.

**Requests**

1. Could you please file a report with SUSE (bugzilla.suse.com or security@suse.de) that the `cpe` in the OES entries has to be percent-encoded, and link the ticket here so we can track it?
2. Please update the PR description: the cause is not "legacy CPE URI rejection" but this one unencoded comma in the OES vendor name, and the number is worth refreshing.
3. Let's record this as a known limitation and as a blocker for retiring `suse-cvrf` — but not for merging this PR.

DmitriyLewen (commented):
Hello @manojkrishnanomula 
Left comments.

Can you also update PR description after changes?

suse/csaf/csaf_test.go
manojkrishnanomula: done

PR #459, opened 2026-06-18
fix(nvd): retry transient body-read errors and 524 responses

DmitriyLewen:
## Description

The NVD updater could fail an entire run on transient errors that were not covered by the retry loop:

- The response body was read **outside** `fetchURL`'s retry loop, so an HTTP/2 `INTERNAL_ERROR` (the stream aborted mid-body after a `200 OK`) was fatal instead of retried.
- Cloudflare's non-standard `524` (origin timeout) fell into the `default` branch and returned an "unexpected status code" error.

This PR reads the body inside the retry loop and adds `524` to the retryable statuses.
A single request attempt is extracted into a `doRequest` helper that closes the response body via `defer`.

## Context

On 2026-06-17 NVD rolled out a schema/metric expansion (SSVC data from CISA-ADP in addition to CVSS, plus an "affected" block per the CVE Record Format).
Every updated record received a new `Last Modified` date, which means a large share of the database was re-stamped in a single day.

As a result, the updater's per-day window started returning almost the whole database at once.
Compare a normal day to the affected day.

Successful run ([run 27657694905](https://github.com/aquasecurity/vuln-list-update/actions/runs/27657694905/job/81795312170#step:7:29)):

```
2026/06/17 00:39:15 INFO Fetched NVD entries total=169 start_index=0
```

Failing run ([run 27744143884](https://github.com/aquasecurity/vuln-list-update/actions/runs/27744143884/job/82078211053#step:7:28)):

```
2026/06/18 07:38:54 INFO Fetched NVD entries total=261624 start_index=0
```

Total number of records currently stored:

```
➜  api find . -type f | wc -l
  357760
```

So a single day's window covered `261624 / 357760 ≈ 73%` of all records — roughly 131 pages of 2000 entries.
Fetching that many pages takes 40–60 minutes, runs into the NVD rate limit, and eventually hits a transient server error (`503`, `524`, or an HTTP/2 `INTERNAL_ERROR`).
Before this change such a transient error failed the whole run and reverted it, so `lastUpdatedDate` never advanced and the next run pulled an even larger window — a self-sustaining loop.

This is not a regression on our side; it is the fallout of the NVD update.
Making the transient errors retryable lets a run survive the spike and finish, after which the window shrinks back to normal.

## Tests

- `TestUpdate` — added two `524` cases: a happy path that succeeds after one reconnect (`retry: 1`), and a sad path that exhausts retries (confirming `524` is now retried instead of returning "unexpected status code").
- `TestUpdate_RetryOnBodyReadError` — hijacks the connection to send a `200 OK` with a truncated body (declared `Content-Length` larger than the bytes sent, then closes the connection), forcing an `io.ReadAll` error after the status code, and verifies the next attempt succeeds.

### DIFF
--- a/nvd/nvd.go
+++ b/nvd/nvd.go
@@ -23,6 +23,11 @@ const (
 	maxResultsPerPage = 2000
 	retryAfter        = 30 * time.Second
 	apiKeyEnvName     = "NVD_API_KEY"
+
+	// statusOriginTimeout is a non-standard Cloudflare status code (524) returned
+	// when the origin server (NVD backend behind the proxy) does not respond in time.
+	// There is no constant for it in net/http.
+	statusOriginTimeout = 524
 )
 
 type Option func(*Updater)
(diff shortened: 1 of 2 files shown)


DmitriyLewen:
test run in fork - https://github.com/DmitriyLewen/vuln-list-update/actions/runs/27763761063/job/82144770085

nvd/nvd.go
```diff
@@ -135,66 +140,86 @@ func (u Updater) saveEntry(interval TimeInterval, startIndex int) (int, error) {
...
-			// But if they start setting a non-zero value, we can use that duration.
-			if headerRetry := resp.Header.Get("Retry-After"); headerRetry != "0" {
-				hRetry, err := time.ParseDuration(headerRetry)
-				if err == nil {
-					ra = hRetry
-				}
+// doRequest performs a single NVD request attempt and closes the response body
+// before returning. It returns the response body on success. A nil body with a
+// nil error means the request should be retried after waiting for `wait`.
+func (u Updater) doRequest(c *http.Client, url string, attempt int) (body []byte, wait time.Duration, err error) {
```
knqyf263: nit: passing `attempt` into `doRequest` leaks the backoff responsibility into a per-attempt function. `attempt` is loop state owned by `fetchURL`, and it's only used here to compute the linear backoff `time.Duration(attempt) * time.Second`.

The only thing `doRequest` really needs to hand back to the loop is the server-mandated wait (the rate-limit `Retry-After`); the backoff itself is the loop's policy. One way to express that is to drop `attempt` and signal "retryable" via a wrapped error that also carries the optional retry-after, so `fetchURL` owns the backoff:

```go
// errRetry signals fetchURL to retry. retryAfter is the server-mandated
// minimum wait (rate limit); zero means apply the caller's backoff.
type errRetry struct{ retryAfter time.Duration }

func (errRetry) Error() string { return "retryable" }

func (u Updater) fetchURL(url string) ([]byte, error) {
	var c http.Client
	for i := 0; i <= u.retry; i++ {
		body, err := u.doRequest(&c, url)
		var re errRetry
		switch {
		case err == nil:
			return body, nil
		case errors.As(err, &re):
			wait := re.retryAfter
			if wait == 0 {
				wait = time.Duration(i) * time.Second
			}
			time.Sleep(wait)
		default:
			return nil, err
		}
	}
	return nil, xerrors.Errorf("unable to fetch url. Retry limit exceeded.")
}
```

`doRequest` then returns `([]byte, error)`: `errRetry{retryAfter: ra}` for 403/429, `errRetry{}` for 503/524/timeout/network/body-read errors, the body for 200, and a plain error for the unexpected-status case. As a bonus this also removes the `body != nil` retry sentinel, since success is now distinguished by `err == nil`.

That said, adding a dedicated error type might be overkill for this single spot, so I'll leave the call to you. Not blocking.

knqyf263 (dismissed):

nvd/nvd.go
DmitriyLewen: Good idea — that's more accurate.

Updated in https://github.com/aquasecurity/vuln-list-update/pull/459/commits/38049fd986a30945de82044b46f6aa55bb1d7660.

New test run from fork - https://github.com/DmitriyLewen/vuln-list-update/actions/runs/27869235304/job/82478367450

knqyf263 (approved):

PR #464, opened 2026-07-03
ci: replace Microsoft Teams failure alerts with Slack

DmitriyLewen:
Migrate the CI failure notifications to Slack.

## What changed

Replace the `Skitionek/notify-microsoft-teams` step with the official `slackapi/slack-github-action` (pinned by commit SHA) in the `nvd`, `redhat`, `debian` and `update` workflows.
Each alert fires on `failure()` and posts the failing workflow name and a link to the run via an incoming webhook.

## Ops follow-up

A new `TRIVY_SLACK_WEBHOOK` repository secret must be added (Slack incoming webhook URL).
The old `TRIVY_MSTEAMS_WEBHOOK` secret can be removed once Slack alerts are confirmed working.

## Testing

Verified on a fork branch — the `nvd` and `update` workflows were forced to fail, and the Slack notification step posted successfully to a test channel:

- [`Update vuln-list-nvd repo` run](https://github.com/DmitriyLewen/vuln-list-update/actions/runs/28660067899)
- [`Update vuln-list repo` run](https://github.com/DmitriyLewen/vuln-list-update/actions/runs/28660069082)

_Slack alerts in the test channel:_
изображение

### DIFF
--- a/.github/workflows/debian.yml
+++ b/.github/workflows/debian.yml
@@ -49,11 +49,11 @@ jobs:
         repository: ${{ env.VULN_LIST_DIR }}
         commit-message: "Debian Security Bug Tracker"
 
-    - name: Microsoft Teams Notification
-      uses: Skitionek/notify-microsoft-teams@e7a2493ac87dad8aa7a62f079f295e54ff511d88
+    - name: Slack Notification
+      uses: slackapi/slack-github-action@45a88b9581bfab2566dc881e2cd66d334e621e2c # v3.0.3
       if: failure()
       with:
-        webhook_url: ${{ secrets.TRIVY_MSTEAMS_WEBHOOK }}
-        needs: ${{ toJson(needs) }}
-        job: ${{ toJson(job) }}
-        steps: ${{ toJson(steps) }}
+        webhook: ${{ secrets.TRIVY_SLACK_WEBHOOK }}
+        webhook-type: incoming-webhook
+        payload: |
+          text: ":x: *${{ github.workflow }}* workflow failed — <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|view run>"
--- a/.github/workflows/nvd.yml
+++ b/.github/workflows/nvd.yml
@@ -49,11 +49,11 @@ jobs:
         repository: ${{ env.VULN_LIST_DIR }}
         commit-message: "NVD"
 
-    - name: Microsoft Teams Notification
-      uses: Skitionek/notify-microsoft-teams@e7a2493ac87dad8aa7a62f079f295e54ff511d88
+    - name: Slack Notification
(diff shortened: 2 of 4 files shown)


.github/workflows/debian.yml
```diff
@@ -49,11 +49,11 @@ jobs:
...
       with:
-        webhook_url: ${{ secrets.TRIVY_MSTEAMS_WEBHOOK }}
-        needs: ${{ toJson(needs) }}
-        job: ${{ toJson(job) }}
-        steps: ${{ toJson(steps) }}
\ No newline at end of file
+        webhook: ${{ secrets.TRIVY_SLACK_WEBHOOK }}
+        webhook-type: incoming-webhook
+        payload: |
+          text: ":x: *${{ github.workflow }}* workflow failed — <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|view run>"
```
nikpivkin: Suggestion: since this step runs with `if: failure()` inside a specific job, it's that job that failed — not necessarily the whole workflow. Using `github.job` here would be more accurate; keeping `github.workflow` alongside it for context is fine.
```suggestion
          text: ":x: Job *${{ github.job }}* failed in workflow `${{ github.workflow }}` — <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|view run>"
```

nikpivkin: nit: missing newline at end of this file or the next one

nikpivkin (changes requested):
I left a couple of small comments

.github/workflows/debian.yml
DmitriyLewen: On `github.job` — it resolves to the **`job_id`** (the map key under `jobs:`), which is `update` in all four of these workflows, not the job's `name:`. There's no context that exposes a job's display name: `github.job` is the id, and the `job` context only carries `status`/`container`/`services`. So `${{ github.job }}` would render the literal `update` in every alert — identical across all four and non-distinguishing.

`github.workflow` instead gives a unique, human-readable name per source (`Update vuln-list-nvd repo`, `Update vuln-list-debian repo`, …), and since each of these workflows has a single job, the workflow-level message already points at the failing job. So I'd prefer to keep `github.workflow` here — happy to revisit if you still see value in the job id.

Refs: [`github.job` = `job_id`](https://docs.github.com/en/actions/learn-github-actions/contexts) · [`job_id` is the unique key, distinct from `name`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id)

That said, your point holds in the general case — I'll keep it in mind for other repos that do have multiple jobs per workflow.

DmitriyLewen: added in https://github.com/aquasecurity/vuln-list-update/pull/464/commits/7e4245bffd89bf673b3a703917db28d4719f2781

nikpivkin (approved):

knqyf263 (approved):

PR #466, opened 2026-07-03
fix(ubuntu): prevent index-out-of-range panic in UpstreamLinks loop

arpitjain099:
The UpstreamLinks loop in `ubuntu.parse()` guards with `j < len(lines)` but reads `lines[j+1]` in the same condition, so once `j` reaches the last line the guard indexes `lines[len(lines)]` and panics with `index out of range`. It triggers on a CVE file whose final line is an `upstream:` continuation of a `Patches_` block with no trailing newline. Since `parse()` runs over the whole ubuntu-cve-tracker via `filepath.Walk`, a single such file would abort the entire Ubuntu update run, so I figured it was worth fixing defensively. The change just switches the guard to `j+1 < len(lines)`, which is what the three other loops in this file already do.

Before: the added test panics at ubuntu.go:335 with `index out of range [3] with length 3`. After: `go test ./ubuntu/...` and `go vet ./ubuntu/` both pass, including the existing subtests.

There's no associated issue since I came across this while reading the parser rather than hitting it in the wild. Happy to open a short bug issue if you'd prefer one on file.


DmitriyLewen (approved):
Thanks!

PR #475, opened 2026-07-29
feat(chainguard): add the Chainguard OSV v3 feed as the chainguard-osv target

plerion-sina:
## Description

Chainguard's secdb feed (`security.json`) is deprecated. This adds a `chainguard-osv` target that mirrors [Chainguard's OSV v3 feed](https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/osv_v3_feed.md) instead, for both the Chainguard and Wolfi ecosystems.

The v3 feed publishes one file per advisory, roughly 837k of them, so this consumes the tar archive of the whole feed rather than the ~837k individual record endpoints. It streams the archive, groups the advisories by package, and writes one file per package. Nothing is extracted to disk.

Output, from a full run against the live feed:

```
vuln-list/chainguard-osv/v3/chainguard/   10,585 files
vuln-list/chainguard-osv/v3/wolfi/         3,623 files
```

That is 362MB on disk, 27.8MB compressed, from 836,784 advisory records. Runtime is 19s with ~1.1GB peak RSS.

Each advisory keeps its CGA ID, its upstream IDs, the architecture from the package PURL, the OSV range events verbatim, and Chainguard's resolution status. Combining the advisories of a package is deliberately left to the consumer, since that needs APK version comparison, which belongs in trivy-db. The OSV schema types are reused from the existing `osv` package rather than redeclared.

Two things are dropped on purpose: each advisory's `modified` timestamp, because no consumer reads it and keeping it would rewrite most of these files on every run, and the severity and reference metadata, because Trivy takes severity for these distros from NVD today.

The existing secdb `chainguard` target is untouched.

Note the output goes to `vuln-list/chainguard-osv/`, not `vuln-list/chainguard/v3/`. It has to: the secdb `chainguard` target does `RemoveAll("vuln-list/chainguard")` on every run, and since each target commits and pushes independently, writing inside it would mean every scheduled run pushed a commit deleting all 14,208 files and then pushed them back.

## Two things worth deciding before this merges

**Where the mirror lives.** 362MB is a lot to add to `vuln-list`. The precedent for a source this size is its own repo, the way `vuln-list-redhat`, `vuln-list-nvd` and `vuln-list-debian` work, and this addition is bigger than all of `vuln-list-debian`. Pointing it at a new `vuln-list-chainguard` is a one line change to the workflow input if you would rather. The monthly squash workflow already handles the daily churn either way.

**Whether it should go through a mirror at all.** trivy-db already fetches nine sources directly into its cache dir without a vuln-list mirror: `ghsa`, `bitnami-vulndb`, `govulndb`, `k8s-cve-feed`, `julia`, `cocoapods-specs`, `nodejs-security-wg`, `php-security-advisories`, `ruby-advisory-db`. The v3 feed is a single tarball at a stable URL, so it fits that shape, and going that route would drop this PR entirely along with the size question above.

The argument against is consistency: all nine of those are language ecosystems, and every OS package source goes through `vuln-list`. Chainguard is an OS source, which is why this follows that convention. The cost of the direct route is that trivy-db would download 250MB and extract 4.2GB across 837k files on every DB build, where this pays that cost once per feed refresh.

Happy to rework it either way, just say which.

## Related PRs

- [ ] https://github.com/aquasecurity/trivy-db/pull/686
- [ ] https://github.com/aquasecurity/trivy/pull/11015

This one needs to merge first, and the target needs to have run at least once, before the trivy-db PR can merge. trivy-db's `Update()` fails the whole DB build if the directory is missing, which is the right behaviour but makes the ordering strict.

## Checklist

- [x] I've read the [guidelines for contributing](https://trivy.dev/docs/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/docs/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/vuln-list-update/blob/main/README.md) with the relevant information (if needed).


plerion-sina:
Closing this one. We're going to carry these changes downstream rather than upstream for now.

The branch stays up if anyone wants to pick it up later, and the description above has the full write-up: what changes between secdb and the v3 feed, the aggregation rules, and the validation against Chainguard's `scanner-audit` across all 10 test cases and both distro variants.

Thanks for the project.

PR #479, opened 2026-08-31
fix(redhat/csaf): remove advisories deleted upstream

nikpivkin:
The mirror keeps 268,248 advisories that Red Hat has already deleted. The delta update only looks at deletions.csv within a short window, and an entry can appear there long after its own timestamp, so whatever is missed once is never reconsidered. The archive is the only view Red Hat publishes of the current state, so every run now checks `archive_latest.txt` and reconciles the local data with a newer archive when one shows up, dropping what is no longer published.

Related PRs:
- [fix(redhat/csaf): drop unused fields to fit GitHub's file size limit](https://github.com/aquasecurity/vuln-list-update/commit/3adf09021669f831270896dd879d16d7e9aea6df)


nikpivkin:
Test run: https://github.com/nikpivkin/vuln-list-update/actions/runs/33386670672/job/99470747155

The reconciliation itself works. csaf-vex went from 19.78 GB in 333,682 files down to 15 GB in 66,122 files, so about 267k advisories that Red Hat no longer publishes were removed. The run also included the field trimming from #477, which is why the size drops more than the file count alone would explain. The extra ~680 files over the 65,440 in the archive are the ones the delta pulled from changes.csv after the snapshot date.

The push failed:

```console
error: RPC failed; curl 92 HTTP/2 stream 3 was not closed cleanly: CANCEL (err 8)
send-pack: unexpected disconnect while reading sideband packet
fatal: the remote end hung up unexpectedly
```

This is a transport error, not a clear rejection because the pack is too large, so I cannot tell yet whether we hit a GitHub limit or just failed to send that much data at once.

Worth separating two cases here. The first push happens only once: it removes 267k files and rewrites everything the archive carries. If it does not go through in one piece, the answer is to split it into several smaller commits.

Every later push should be much smaller, since a weekly reconciliation only commits what actually changed. I cannot measure that until the first push lands, so it is the next thing to sort out.

### The new feed

#425 migrates this fetcher to the new vex-feed format, and https://issues.redhat.com/browse/SECDATA-1285 discusses how that feed is meant to be consumed. Two answers from Red Hat there matter for this PR. The new archive is generated once a day instead of once a week, and they say they did not expect consumers to build incremental updates on top of the archive at all.

If we follow that model after the migration, most of this PR is not needed. The reconciliation, the cursor and the time buffer only exist to patch up the delta model, and re-fetching a daily archive removes the problem instead of patching it. Keeping this as a draft until we decide.

PR #482, opened 2026-09-21
fix(redhat): raise Security Data API per-year fetch timeout to 40m

DmitriyLewen:
## Description

`retrieveRedhatCveDetails` fetches all CVEs of a single year within one 30m budget.
The number of CVEs per year keeps growing: ~6.1k in 2024, ~7.9k in 2025 and ~14k in 2026 so far.
At the usual rate of ~8.4 CVE/s the current year already takes 27m34s of the 30m budget, so we are close to the point where a year stops fitting and the whole update fails and gets reverted.

This PR raises the per-year timeout 30m → 40m and moves it into a named constant.


knqyf263:
Looking at the failed run, fetching 2,553 of 7,869 CVEs took about 30 minutes. At the same average rate, fetching all of them would take roughly 92 minutes, so increasing the timeout to 40 minutes would still fall well short.

Is the intent of this change to give healthy runs, which currently take around 27 minutes for 2026, some extra headroom, rather than to make runs experiencing the reported slowdown succeed?

DmitriyLewen:
Yes, you are right — 40m would not have saved a run that hit this slowdown: at 1.4 CVE/s the 2025 batch alone needs ~92 minutes.

What I found while investigating the failures is a different problem: we are approaching a limit past which every run will fail regardless of the API state.
Year 2026 already holds ~14k CVEs and takes 27m34s of the 30m budget in a healthy run (~8.4 CVE/s), and the number of CVEs per year keeps growing.

As for the failures themselves — they are 2 consecutive ones out of 12 runs over the last three days, and the neighbouring runs fetched the same years about 6x faster.
So I assume this is temporary degradation on the Red Hat side.
I will keep watching it and move on to solving the problem if it happens again.

Re-reading the PR description, I agree it was misleading about what this change does, so I rewrote it.

knqyf263 (approved):

